PR #64, opened 2015-01-07
Blob Replication

dominictarr:
This gives us a complete first pass at attachments.
When an instance sees a message containing a ext link, it requests it from peers,
with out requesting from too many peers at once, etc.

There is still more work to do here, but this is enough to do a photo sharing app, for example.

The most important things to do next is to write a new stream encoding that handles json and binary.
It can keep newline delimited frames for json, but needs length delimited frames for binary, mainly for attachments.

And then, ext links should have a length too, and peers shouldn't request ext links that are too large, and if an blob is sent that _is_ too long, then reject it... this will enforce the length limitation and prevent too large objects from polluting the network.


bin.js
```diff
@@ -147,7 +153,51 @@ function next (data) {
...
+            rpc.blobs.add(function (err) {
+              if(err) return next(err)
+              link.ext = hasher.digest
+              next()
+            })
+          )
+        }
+      })
+
+      function next () {
```
pfraze: this is missing its err param

pfraze:
:+1: this is looking pretty good!

bin.js
dominictarr: fixed in 391b072bd14d1d7874ae2ebec25339ab798f8593

pfraze:
added some logging, called blobs.want from the cli and the blobs plugin flooded grimwire (only reachable peer) with the has() request

what strategy should we use for that? we do want to do repeat polling to a peer (i think) because they may receive the blob after an initial has() request, but we dont want floods. maybe ask each peer 1/minute?

more generally (you've put more thought into this so feel free to disagree) i'm unsure about auto downloading attachments. i worry it could run away in the background

what about an on-demand client-request style flow, where a want() call puts out the has() request to all current peers, returns false (and forgets about it) if none has, downloads then returns true if any has? then a client calls get() as usual. you'd have to wait the first time you want a file but thats no worse than the existing web. on the plus side, it'd be client-initiated

dominictarr:
by flooded you mean request over and over? that is certainly a bug. can we make a test for this bug?

I think we probably want some degree of background replication, or some way to add things to a whitelist, or for apps to beable to request things they will need.

Wether we want this to be background or foreground is part of a larger dicussion, i think.
there are some attachments you'd want to only download after user's request (generally: large things)
but others that you'd want to download automatically.

For example, say we make a packagae manager on top of ssb - then if I use a lot of modules by Alice, then I might want those all just replicated, incase I need one later (packages are pretty small) and it would also serve to ensure the packages I want are widely available.

I measured a while back, and I had 600 packages in my npm cache, from 300 authors. replicating all the modules from the authors who wrote those wouldn't be that much data... but replicating _all_ the data would be.

however, i figure, whatever we end up using will be a subset of replicating everything, so replicating everything would be a good first stage, and then filter it somehow.

pfraze:
> however, i figure, whatever we end up using will be a subset of replicating everything, so replicating everything would be a good first stage, and then filter it somehow.

that makes sense. it also occurred to me this morning that on-demand (as i described it) wouldnt work with pub servers since they have no client making requests for attachments

this is tricky like the pub-messages thing is tricky. i'm going to think about it and write more later

pfraze:
(I'll get a test committed in a bit)

dominictarr:
okay lets treat this as a tracking pull request for now, I'm just working on having sizes on the ext links.

dominictarr:
@pfraze re: polling if peer does not have a thing, maybe there is an event emitted when blobs are added in realtime? then connected peers can see that, and request that blob, currently, it wouldn't request that blob until it connected again.

@pfraze also, can you clarify what you mean by "flooding"? do you mean requesting over and over in a loop?

pfraze:
`has` events would be more efficient but could be missed. we're getting into the realm of DHTs here -- do we want to consider one of those algorithms?

> also, can you clarify what you mean by "flooding"? do you mean requesting over and over in a loop?

yes, with very little delay

dominictarr:
If you have requests and events then they should never get missed - you could call `has` and get `false` but then if the peer receives that hash while you are connected to them, you'd see the event and know you can request it now.

Okay I better fix flooding before we merge this.

dominictarr:
okay this also needs upgrading to secure-scuttlebutt@8

dominictarr:
okay I have confirmed the flooding behavior.

dominictarr:
okay I have a fix, just gotta make the tests fail properly ;)

dominictarr:
Okay, flooding should now be fixed. Peers request blobs, but now remember that they heard back "no", and do not check that hash again as long as they are on the same connection.

Oh, I better add "has" events, else peers might wait a while to request a recently published file.

dominictarr:
Okay, this is ready!

pfraze:
awesome looks :+1:

dominictarr:
can we merge this?

pfraze:
yes

pfraze:
published `2.1.0`

PR #65, opened 2015-01-07
add trust and flag edges to friend graphs

pfraze:
modifies the friend plugin functions to optionally take the desired graph

```js
server.friends.all('follow')
server.friends.all() // same as ^
server.friends.all('trust')
server.friends.all('flag')

server.friends.hops(alice, 'follow')
server.friends.hops(alice, 'trust')
server.friends.hops(alice, 'flag')
```

also adds edge-deletion to the graph construction

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "multiblob": "~1.2.0",
     "bash-color": "~0.0.3",
     "pull-inactivity": "~2.1.1",
-    "graphmitter": "~1.1.2",
+    "graphmitter": "~1.3.0",
     "non-private-ip": "~1.1.0",
     "ip": "~0.3.2"
   },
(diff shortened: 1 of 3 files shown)


pfraze:
published `1.2.1`

dominictarr:
it would be better to have the graphs in an object and then just pick one off than to have the `if...else`

``` js
var graphs = {
  trust: new Graphmitter(),
  follow: new Graphmitter(),
  flag: new Graphmitter()
}

//and then. etc
if(!Object.hasOwnProperty(graphs, type))
  throw new Error('no graph'))
return graphs[type].toJSON()
```

It does the same thing, but is much more terse, and has less edge cases, because it doesn't support the plugal version. There is no benefit to having multilpe names to the same thing unless we are supporting a legacy pattern that hasn't been completely cleaned up yet.

pfraze:
ok :+1:

pfraze:
we're going to need to re-approach this per https://github.com/ssbc/phoenix/issues/257

PR #67, opened 2015-01-09
added peer backoff on successive failures

pfraze:
pretty simple algorithm, updates the peer selection to only choose a peer if `Math.random() < (1 / (1+e.failure))`


dominictarr:
:+1:

Issue #72, opened 2015-01-15
invite code allows too many uses


pfraze:
two bugs with invite code use

 1. one user can use an invite code multiple times. we should check if sbot is already following and, if so, return success without actually using the code
 2. i created an invite code with 1 use, but it's allowed me to use it multiple times. not sure why

dominictarr:
are you testing this manually or in a script?
I admit there may be a racecondition if you request invites really fast... hmm, that is an attack vector that should be plugged.

pfraze:
manually in the phoenix ui

dominictarr:
okay that shouldn't happen. I guess we should have a way to check how many follows have been used, too.

PR #74, opened 2015-01-16
invite code fixes

pfraze:
closes https://github.com/ssbc/scuttlebot/issues/72

decided to make 'already followed' an error condition so that the client could reason about it


pfraze:
published `2.0.1`

PR #80, opened 2015-01-21
Blob tooling (logs, blobs.wants func)

pfraze:
some additions to help us debug/monitor the blobs exchange

adds logging and the `blob.wants` function, which gives the current jobs queue


dominictarr:
yes, this is good.

pfraze:
if it passed tests! debugging...

PR #92, opened 2015-01-23
fix: set home path correctly on windows

pfraze:


pfraze:
published `2.1.7`

dominictarr:
maybe it would be better just to use the [osenv](https://github.com/isaacs/osenv) module (used by npm) and not have to inline this code.

pfraze:
ok

PR #99, opened 2015-01-27
added gossip.connect rpc method

pfraze:
new rpc function in the gossip plugin lets us trigger syncs manually, eg from phoenix's address page

decided to implement it, for now, such that it only connects to known peers. however, we could improve it to allow previously-unknown peers if we feel that's useful

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "ssb-keys": "~0.4.1",
     "ssb-msgs": "~2.0.0",
     "ssb-msg-schemas": "~1.0.0",
-    "ssbplug-phoenix": "~2.2.6",
+    "ssbplug-phoenix": "~2.2.10",
     "stream-to-pull-stream": "~1.6.0",
     "explain-error": "~1.0.1",
     "map-merge": "~1.1.0",
--- a/plugins/gossip.js
+++ b/plugins/gossip.js
@@ -43,6 +43,7 @@ module.exports = {
   manifest: {
     seeds: 'async',
     peers: 'sync',
+    connect: 'async'
   },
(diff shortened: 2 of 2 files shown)


pfraze:
published `2.1.9`

dominictarr:
what happens if the user tries to automatically connect to everything?

pfraze:
they'd be able to. i am noticing, however, `connectTo` doesnt check if the peer is already connected. need to add that, https://github.com/ssbc/scuttlebot/issues/101

dominictarr:
hmm, I think this is an argument for rolling this into the other code that connects clients... connectTo could just add a high priority connect job, and then that is processed with the main connecting code path.

pfraze:
i considered having connectTo bump priority instead of starting a connect, but i figured it was nicer ux if it just acted on-demand. we can make the ux work for higher priority though (have its status line say "waiting to connect..." or something like that).

why are you wanting to guard against the user auto-connecting to everything? trying to make sure peers are good citizens?

dominictarr:
I think I am trying to maintain code complexity... also with a mind to the underlying "jobs" abstraction, we'll probably want to be able to create a blobs job to download a particular blob with high priority too.

pfraze:
yeah, makes sense

dominictarr:
It's just my intuition I guess... there is a jobs concept that is used to manage who to connect to, and it gets more complex if you add another way of connecting... but if you just connect via the same mechanism for a special reason, that is still simple.

Stopping the user from requesting too many connections is probably not a likely problem... but it may be for blob replication.

PR #102, opened 2015-01-31
bin.js now correctly connects to localhost

pfraze:
`bin.js` was timing out when i attempted to connect to it. this seemed to solve it.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -109,7 +109,7 @@ var type = get(manifest, cmd)
 
 if(!type) return usage()
 
-var rpc = require('./client')(config, manifest, function (err) {
+var rpc = require('./client')({ host: 'localhost', port: config.port }, manifest, function (err) {
     if(err) throw err
   })
 


pfraze:
published `2.1.11`

dominictarr:
This is a breaking change, because now you cannot connect to a remote host that you are the master of.
eg, to create an invite code i do `ssb_host=<ip> ./bin.js invite.create 100`
and it connects to the scuttlebot on that ip, not on local host.
To be able to do this, you have to set the `"master": hash(pubkey)` in the remote scuttlebot's config.

pfraze:
oh, my bad. I wasn't aware of that.

hmm, ok. it'd be pretty clean if we could keep using the `host` env var. maybe we update `client.js` to do

```js
var nonPrivate = require('non-private-ip')
//...
var host = (config.host == nonPrivate()) ? 'localhost' : config.host
var rpc = require('./client')({ host: host, port: config.port }, manifest, function (err) {
  if(err) throw err
  // ...
```

yeah?

dominictarr:
hang on, client.js already defaults host to `localhost` what does your config look like? do you have another `host` property configured?

pfraze:
prior to this change, `bin.js` was passing the object exported by `config.js` to the `client.js` function. the `host` value of `config.js` is set by `non-private-ip`, which, for my laptop, was an ipv6 address -- and that was failing

dominictarr:
so, is your laptop publically addressable by that ipv6 address?
ip addresses have a private range, and non-private-ip checks for that.
my ipv6 address is `fe80::5a4:f331:d041:9f71` which is private (starts with `fe80`)

If your laptop has a publically addressable ipv6, that is pretty cool.

pfraze:
Interesting, didn't know about the private range.

This is what `non-private-ip` gives me on my laptop: `2605:6000:e900:d200:2ef0:eeff:fe01:839c`. If I google "whats my ip" they tell me `2605:6000:e900:d200:f4bb:ab57:1a39:ac9`.

Here's what `os.networkInterfaces()` gives me:

```js
{ lo0: 
   [ { address: '::1',
       family: 'IPv6',
       internal: true },
     { address: '127.0.0.1',
       family: 'IPv4',
       internal: true },
     { address: 'fe80::1',
       family: 'IPv6',
       internal: true } ],
  en0: 
   [ { address: 'fe80::2ef0:eeff:fe01:839c',
       family: 'IPv6',
       internal: false },
     { address: '2605:6000:e900:d200:2ef0:eeff:fe01:839c',
       family: 'IPv6',
       internal: false },
     { address: '2605:6000:e900:d200:f4bb:ab57:1a39:ac9',
       family: 'IPv6',
       internal: false },
     { address: '192.168.0.2',
       family: 'IPv4',
       internal: false } ],
  awdl0: 
   [ { address: 'fe80::184a:a8ff:fe2d:54ce',
       family: 'IPv6',
       internal: false } ] }
```

Looking at my OS TCP/IP settings, I see my router has an IPv6 address as well. I don't know what the `...:839c` address represents -- is it possible that's my internal LAN address, or would that be in the `fe80:..` range as well?

dominictarr:
heh, well it looks like if we choose ac9 instead then your laptop, on your home network could be a pub server.

dominictarr:
There are so many ipv6 addresses that they don't assign you one, they give you a whole bunch.
So, maybe you just have two addresses you can use? maybe they will both work?
However, since ":" is used to separate the port, my guess is that we are breaking that by just appending `:<port>`

here is what the url module does when you give it a ipv6

``` js
 url.format({hostname: <ipv6>, port: 2000})
'//[fe80::5a4:f331:d041:9f71]:2000'
```

PR #106, opened 2015-02-06
refactor out configuration

dominictarr:
Applications that connect to scuttlebot as a client need to have the configuration so that they can find the manifest. This module means they do not have to depend on scuttlebot... probably we refactor out ./client.js too...
which could contain the default manifest...


pfraze:
looks good. we should still publish a `ssb-manifest` module i think

pfraze:
published `2.2.1`

PR #112, opened 2015-02-08
manage plugin http use with connect.

dominictarr:
This change was necessary to make my dynamic-manifest pull request to phoenix.

The main upside of this change is it enables more than one plugin to add http handlers,
Also, it serves the manifest (as either json or js) and the blobs plugin hosts hosts blobs.

Ultimately, this means that the manifest only needs to be maintained in one place.


pfraze:
i dig it

pfraze:
published `2.3.0`

PR #115, opened 2015-02-09
use config directly in client

dominictarr:
this reverts https://github.com/ssbc/scuttlebot/pull/102
and should resolve the problem but depends on https://github.com/ssbc/ssb-config/pull/4


pfraze:
published `2.3.2`

Issue #116, opened 2015-02-09
questions on info blob wants and blake hashes


ralphtheninja:
![screenshot from 2015-02-09 23 07 41](https://cloud.githubusercontent.com/assets/308049/6116900/a31bc5d4-b0b0-11e4-8a43-0acfa2b0b77e.png)

1. What does this message mean?
2. Why are there three entries for a blob with the same hash?
3. Why did you pick blake hash and what is it used for?

pfraze:
1. It means scuttlebot is going to search the network for those blobs, and it hasn't found them yet. Sbot scans all received messages for `ext` links -- the links which reference blobs -- and adds them to the search queue. The ones that show during load are the yet-unfound blobs, and there's a good chance they may never be found, but we don't have a "give it up" algorithm yet.
2. There were 3 messages referencing that blob, and the message lets us know each time, even though the effects of wanting a hash are idempotent.
3. We use blake2s for content hashing. We encode the hash in base64 and then append the `.blake2s` tag to identify the algorithm. Blake was chosen by @dominictarr so you'd have to ask him what his criteria for selection were.

dominictarr:
I chose blake2s because it is the fastest hash implementation in javascript.
We use blake2s for everything.

The want log is written when something requests that a blob be retrived:
https://github.com/ssbc/scuttlebot/blob/master/plugins/blobs.js#L254-L263

Since the log is outside of the check it may output it twice, but will only request it once.
We should move it else, so that the log messages are not duplicated.

PR #117, opened 2015-02-10
only emit wanted blob once

ralphtheninja:
Addresses point 2 in https://github.com/ssbc/scuttlebot/issues/116


pfraze:
:+1: so small I'm not going to publish

ralphtheninja:
No worries :)

dominictarr:
:+1

Issue #120, opened 2015-02-11
crash: cat data.json | sbot add --


ralphtheninja:
```
$ cat data.json | sbot add --

/home/lms/src/ssbc/scuttlebot/bin.js:202
          if(err) throw err
                        ^
TypeError: Cannot read property 'type' of null
    at Object.add (/home/lms/src/ssbc/scuttlebot/node_modules/secure-scuttlebutt/feed.js:50:19)
    at Object.add (/home/lms/src/ssbc/scuttlebot/node_modules/cont/node_modules/continuable/to.js:11:28)
    at EventEmitter.api.add (/home/lms/src/ssbc/scuttlebot/lib/api.js:48:10)
    at Object.PacketStream.request (/home/lms/src/ssbc/scuttlebot/node_modules/muxrpc/index.js:79:27)
    at onRequest (/home/lms/src/ssbc/scuttlebot/node_modules/muxrpc/node_modules/packet-stream/index.js:26:12)
    at Object.p.write (/home/lms/src/ssbc/scuttlebot/node_modules/muxrpc/node_modules/packet-stream/index.js:181:33)
    at /home/lms/src/ssbc/scuttlebot/node_modules/muxrpc/pull-weird.js:52:15
    at loop (/home/lms/src/ssbc/scuttlebot/node_modules/muxrpc/node_modules/pull-stream/sinks.js:15:33)
    at /home/lms/src/ssbc/scuttlebot/node_modules/muxrpc/node_modules/pull-stream/throughs.js:97:9
    at /home/lms/src/ssbc/scuttlebot/node_modules/pull-stream/throughs.js:97:9
```

```
$ cat data.json
{
  "type": "sys-stat",
  "text": {
    "os": {
      "type": "Linux",
      "arch": "x64",
      "platform": "linux",
      "loadavg": [0.41845703125,0.587890625,0.6923828125]
    },
    "process": {
      "memory": 40370176,
      "memoryInfo": {
        "rss": 40370176,
        "vsize": 4081443667968
      },
      "cpu": 0,
      "cpuInfo": {
        "pcpu": 0,
        "pcpuUser": 0,
        "pcpuSystem": 0,
        "cpuTime": 0
      }
    }
  }
}
```

ralphtheninja:
I traced some stuff and for some reason `sbot` doesn't read the `--`. I'll try to dig some more.

/cc @pfraze @dominictarr

dominictarr:
confirmed

[...]

PR #121, opened 2015-02-11
enable -- with minimist to fix stdin crash bug

ralphtheninja:
Fixes https://github.com/ssbc/scuttlebot/issues/120

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -57,7 +57,7 @@ function usage () {
   process.exit(1)
 }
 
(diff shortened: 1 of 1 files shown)


dominictarr:
I think I'd rather leave it as standard minimist and just use "."
although, using this minimist feature could be useful so that it's possible to set configuration details on the command line, maybe later. Currently we need to use `ssb_{property}` for this.

ralphtheninja:
Np. I have the '.', so I'm good :)

PR #131, opened 2015-02-24
package.json fixes

ralphtheninja:
Gets rid of this warning

```
$ npm i
npm WARN package.json Dependency 'osenv' exists in both dependencies and devDependencies, using 'osenv@~0.1.0' from dependencies
```

Also `rc` is no longer used


pfraze:
:+1: thanks as always!

PR #134, opened 2015-03-03
device-local program perms: self-signed messages, "local" role, etc

pfraze:
@dominictarr this shows what it takes to get apps to self-sign their messages. Have a look at the diff -- pretty straight-forward.

thoughts/questions
 - Would we want to maintain a function for publishing under the main sbot identity? `addMain` or something? phoenix will need something
 - https://github.com/ssbc/scuttlebot/compare/apps?expand=1#diff-0364f57fbff2fabbe941ed20c328ef1aR16 is a little awkward
 - We'll need a way to authenticate the RPC connection well enough to get `add` access. For this demo I use the main keypair


dominictarr:
I think the answer is to take all this and bunde it into a ssb-client module,
that would handle the createMessage call etc, and handle creating the init message, etc
the user would just give it the key they want to use.

pfraze:
@dominictarr I updated https://github.com/ssbc/ssb-client per your suggestions, and brought it to a working place

I also updated this PR to add the needed functions for anonymous RPC peers (but I'm just now remembering the "local connections only" restriction, so maybe we need another role -- `local`?). Take a look and see if you think any of those shouldn't be there, whatever the rpc role will end up being. I was wondering whether we wanted `whoami()` in there. If the peer must be local, the concern would be websites. Will we want to have sites request access first? If so, giving them `whoami` wouldnt bother me.

dominictarr:
whoami is mainly for debugging. Yeah I think a "local" (which is like anonymous + would be good)...

Oh, so, this actually allows us to have a light weight web client - it would just persist the keys,
but keep everything else on the server. the ssb version of a bitcoin webwallet that does the crypto in js, like coinpunk.com... that is not as good as hosting it locally your self... but if it's possible, people will do that whether I like it or not ;)

dominictarr:
A light weight client could store only your own messages - that way it would avoid rewriting two messages with the same sequence. it wouldn't even have to always connect to the same sbot server, either.

pfraze:
1.
if we add a signing challenge to `whoami`, it can be a way to authenticate users on the web. that'd be rad

-

2.
it's true that having web clients use `ssb-client` means js crypto. How defensible is that? There are 3 operations that need to happen right now: key generation (a random function) prev-hash calculation (a blake2s digest function) and message signing (an ecc sign function). The latter two are verified by scuttlebot, but what about the key generation? Can the browser produce good random bits? And we'll just have to accept that the keys are stored in dom storage, and aren't very secure. Then there's also message encryption, when the time comes for that, which is another barrel of monkeys. 

Hmm. I think maybe web apps should keep authenticating their domain (as phoenix supports now) and have their crypto ops and keystorage done for them. Webapps cant spoof their Origin headers, so we can use that for identification, and I think it'd be possible to protect those stored feeds from local code which CAN spoof the Origin header through a user-authorization flow that gives the app's rpc connection an access token to auth with. If that works, then an attacker would have to either hijack the browser during the auth flow, or read the keyfiles off the disk

-

3.
`ssb-client` can work entirely as a client thanks to the `publish` function, but we can run a full ssb instance in the app for the reasons you suggest. That would look something like:

```js
var ssb = require('secure-scuttlebutt/create')(...)
var keys = require('ssb-keys').generate()
var feed = ssb.createFeed(keys)

var client = require('ssb-client')(keys, 'localhost', function (err) {
  pull(
    ssb.createHistoryStream({id: feed.id, live: true, keys: false}),
    client.createWriteStream()
  )
})

feed.add({ type: 'post', text: 'hey!' })
```

Though sbot doesnt expose `createWriteStream` over RPC right now. If that's for security reasons, we'd have to reapproach the mechanics of this, but that example gives the basic idea

-

4.
How do you feel about `publish` as our horse word? Should we replace `feed.add` with `feed.publish`, ad leave `ssb.add` as the function that takes the fully-formed message?

pfraze:
we discussed item 2 a bit more in IRC, and came to the conclusion that applications were not meaningfully more secure if their keypairs or crypto ops were managed for them, since a compromise event was not significantly different whether it resulted in a stolen app keypair or use of the app's feed via an API

for now, let's proceed with web apps keeping their own keypairs. going to revert the related changes...

pfraze:
@dominictarr you good with everything here? this is the last remaining PR for the new apps system

dominictarr:
sorry, yes lets merge this!

pfraze:
bumping major due to add/publish api change

pfraze:
published `4.0.0`

PR #140, opened 2015-03-10
when ssb-client connects, do not treat it as authed.

dominictarr:
this emits an "unauthorized" event, but, since ssb-client now cb() silently, it does not dump a confusing error either. Without this, since some plugins expect auth's cb to have arguments, scuttlebot was crashing. This fixes that, and makes ssb-client work.


pfraze:
:+1:

pfraze:
published `3.0.2`

PR #142, opened 2015-03-22
make all timeouts use configured timeout

dominictarr:
I changed this while investigating the blobs test problem, I don't think this fixed it, but it is better to have configured timeouts instead of hardcoded timeouts.


pfraze:
ok looks good

dominictarr:
after working with this at my end, I think you'll want to merge https://github.com/ssbc/scuttlebot/pull/134 first

Issue #150, opened 2015-04-04
plugin commands not working (invite.create and invite.addMe at least)


aredridel:
```
:; ./bin.js invite.create 100
sbot {cmd} {options}
```

No matter what commands I use with invite, I just get the terse use instruction.

pfraze:
Sorry for the sorry state of our live help. We have a standing issue to deal with that.

@dominictarr I tried it out and got the same results. My `~/.ssb/manifest.json` is missing the plugin APIs. When the server command is run and that file is written, it appears to happen before the plugins have a chance to register themselves with sbot, and thus they're not included in the manifest file.

Is this possibly new? I haven't noticed plugin commands failing before now.

dominictarr:
confirmed. also happens for `friends.all` and `friends.hops`.
I think I know what this is, startup is now async because of `needsRebuild`,
but bin.js doesn't wait until that is finished.

aredridel:
Aha!

Out of curiosity, what's the purpose of the manifest? Seems a lot like what could be provided statically.

dominictarr:
the manifest is a JSON file that is required on the client (added to browserify bundle) so that the rpc can add those methods to the client, othewise you'd have to call all methods as `call('invite.create', ...)`
I admit it's slightly awkward, but it's the simplest way to do this currently.

dominictarr:
this will be fixed by the above PR

PR #151, opened 2015-04-04
do not write manifest until plugins are loaded

dominictarr:
fixes https://github.com/ssbc/scuttlebot/issues/150

the problem was: https://github.com/ssbc/scuttlebot/issues/150#issuecomment-89483263


aredridel:
That does fix it for me!

pfraze:
:+1: looks good

dominictarr:
merged into 4.0.3

Issue #149, opened 2015-03-31
Attachments 


No9:
@dominictarr I see you are thinking that large attachments should be torrents 
https://github.com/ssbc/scuttlebot/wiki/FAQ#attachments 

Are you working on this? 
I am back in the torrent zone at the moment and this interests me  so I am tempted to have a shot.

pfraze:
It's one of our "gee whiz, what if" ideas, so it's not on our roadmap atm. By all means, give it a go!

@dominictarr feel free to give your thoughts. I think the way to go at it would be to put the magnet URI in an ext link, eg `{ type: 'foo', torrent: { ext: $MAGNET_URI } }`. Then we'd need to update the blobs plugin in sbot so that it doesn't try to fetch that blob (it should look at the `ext` value and discern if it's a content-hash or something else, eg a magnet link). After that, it's just a matter of having some code watch the incoming stream and operate the torrents, which you could do with a custom sbot plugin or with an external program.

Please post your implementation notes/plans as you go! I'd like to learn more about the torrent protocol.

No9:
Thanks for the steer @pfraze  
I'll have a look at your recommendations and see how it plays out. 
I expect a tracker to be put into the public server as well but I will see where we it lands.

dominictarr:
@No9 this would be awesome!

@feross is currently working on the finishing touches to webtorrent (which benefits immensely from having hybrid torrents that can bridge between tcp and webrtc)

One other thing, it would be very good to include the size of the torrent in the link,
then peers can download things preemptively, but not dedicate too much resources.

@pfraze @No9 it might be better to have a slightly different format there,
I suspect that overloading `ext` will add complications later. It might be better to just have a `torrent` type? `{type: 'iso', <rel>: {torrent: <magnet_link>}}` or using an ext link would be ok if it was a link to the torrent file.

pfraze:
Yeah the downside to using `.torrent` is that it won't get indexed by sbot as a link.

dominictarr:
@pfraze we could add that though! there are more subtleties here too, like a magnet link has the info hash (which is deterministic) but also links to trackers (which is easy to change) it would probably be very good to have pub servers be trackers, too.

pfraze:
@dominictarr that's true, I'm just thinking, will more people want to add blob protocols (ipfs, http)? and maybe it'd be better to overload `ext` than to split them all out.

I agree about pubs being trackers, that makes plenty sense.

dominictarr:
[...]

Issue #148, opened 2015-03-29
bin.js commands are failing because it attempts to use public IP for localhost


pfrazee:
I remember dealing with this before, but couldnt find the issue.

The `ssb-config` module is using `non-private-ip` to set the host. On my laptop, this is causing `bin.js` to stall indefinitely on a command. Issues:
- The connect attempt should timeout at some point, right now it doesnt
- I think `bin.js` is meant to default to localhost, so the use of `non-private-ip` is a problem there

dominictarr:
are there steps are reproduce? (like, a particular network it happens on?)

pfraze:
It's occurring on my home network.

I'll need to dig into it a bit to be sure about this, but my guess is --

The IP coming out of `non-private-ip` is for my house, in which case a connect-attempt terminates at my router.

pfraze:
Ah, so I have my mac configured to disable all incoming connections in the software firewall settings. When I turn the firewall off, instead of a long nothing, I get an immediate ECONNREFUSED

pfraze:
Still not sure why that is but it's a change

dominictarr:
right, so the problem is that it's connecting to the local network ip address, instead of localhost?

`non-private-ip` means you have to do one less thing to set up a pub server (it autodetects it's ip when hosting). We changed this one time, forcing the client to connect to localhost,
but that broke setting a remote host and connecting to a remote with `ssb_host=<ip> sbot <cmd>`

hmm, maybe servers should use the non-local ip, but clients should use localhost when you are already on the same machine?

pfraze:
> hmm, maybe servers should use the non-local ip, but clients should use localhost when you are already on the same machine?

yeah, we might consider a `--pub` switch, which might also be useful for enabling the pub-specific frontend (if we choose to go forward with that)

dominictarr:
hmm, that might be simple enough, although if we make the network detection slightly smarter I think we could skip that.

aredridel:
This happens for me, on a machine with IPv6 public IPs. Server listens on 0.0.0.0, bin.js connects on the v6 public.

dominictarr:
do we need two servers to listen on BOTH ip4 and ipv6?

aredridel:
If on linux and `v6only` is set, yes; otherwise, no, listening on v6 will listen on v4 too.

aredridel:
[...]

PR #155, opened 2015-04-06
[Disscussion] torrent tracker on public node 

No9:
Hey All 

As mentioned in https://github.com/ssbc/scuttlebot/issues/149 this is a torrent-tracker for a public server. 

The integration is around here 
https://github.com/ssbc/scuttlebot/compare/master...No9:master#diff-168726dbe96b3ce427e7fedce31bb0bcR335 

When a ```host``` param is supplied then a torrent tracker will be created. 

This is my first integration so if anyone has any suggestions/tips on the way that I have put the pluggin together then please feedback.  

Note the tracker plugin will log on start up:  
```
info SiNB SBOT  torrent-tracker "http server listening on 8000"
info SiNB SBOT  torrent-tracker "udp server listening on 8000
```

If this is too chatty then it can be turned off with ```quiet : true``` 
https://github.com/ssbc/scuttlebot/compare/master...No9:master#diff-1701aeec89893e26031e65ef7176b384R22 

I will also consider the impact of these bittorrent-tracker events when I get more into the ssb code base. 
https://github.com/feross/bittorrent-tracker/blob/master/bin/cmd.js#L69 

I will start looking at the clientside of this now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "scuttlebot",
   "description": "A bot-server for the phoenix network",
-  "version": "4.2.2",
+  "version": "4.2.3",
   "homepage": "https://github.com/pfraze/scuttlebot",
   "repository": {
     "type": "git",


pfraze:
Cool. The `.host` config value is always going to be set (if not explicitly set, it's autoconfigured) so that's not going to be an effective toggle.

No9:
@pfraze thanks for the feedback. 
How do you feel about having a --public flag?

BTW I noticed that when running ```sbot server``` the host property is currently empty.

pfraze:
RE: the config.host, hmm, that may be related to this https://github.com/ssbc/ssb-config/blob/master/index.js#L6. I think we may have a solution for that, suggested at https://github.com/ssbc/scuttlebot/issues/148#issuecomment-89898196

I think probably the right config value to use for this is `.torrent-tracker`

No9:
@pfraze updated with the --torrent-tracker option 
Will track the #148 for updates.

pfraze:
:+1:

plugins/torrent-tracker.js
```diff
@@ -0,0 +1,34 @@
...
+function isString (s) {
+  return 'string' === typeof s
+}
+
+exports.name = 'torrent-tracker'
+exports.version = '1.0.0'
+
+//trust-proxy 'x-forwarded-for' header from reverse proxy 
+//tell clients to announce on this interval (ms) 
+exports.manifest = {
```
pfraze: Oh, you know, this isnt how the manifest should be used. This looks like you're using it for config. The manifest is actually a listing of the methods in the plugin which are RPC-accessable, in the form:
```
{
   methodName: type
}
```
Where `type` can be `'sync'`, `'async'`, `'source'`, `'sink'`, and `'duplex'`. We need to document those, but here it is briefly:

- async: last param is a `cb(err, ...)`
- sync: constructed like a typical sync function, which is handy if called locally by sbot, but the rpc will convert it into an async behavior. If it throws, will `cb(exception)`. If it returns, will `cb(null, returnvalue)`
 - the other three: pull-stream types

You dont have any methods yet, so the manifest should just be empty.

@dominictarr where should plugin config go? For now, I think just stick it on some other object.

dominictarr: well how I did it in other places was take a sub config object off the global config
https://github.com/ssbc/scuttlebot/blob/master/plugins/gossip.js#L56
and then have inline defaults https://github.com/ssbc/scuttlebot/blob/master/plugins/gossip.js#L160

that does make it a bit difficult to see what the settings that plugin would use.
If we have a lot of plugins with config it might be better to have a config section on the plugin.
But for now it might be simplest to take the subobject, and then set defaults on it.

dominictarr:
it should probably have a bit of API so you can query it's status over sbot api.

What would you need to use this? I mean, to actually get a torrent?

No9:
@dominictarr which one of the plug-ins would be best to look at for the status query scenario? 

RE: Trackers 
Basically a tracker keeps a list of who has got the actual content of a torrent. 
When a client opens a .torrent file it contains one or more trackers.

The client then asks the tracker(s) for more information on which peers have the torrent content. 
https://en.wikipedia.org/wiki/BitTorrent_tracker

We don't necessarily have to carry this.

We could use some of the open trackers: 
 
```
udp://tracker.publicbt.com:80
udp://tracker.openbittorrent.com:80
udp://open.demonii.com:1337
udp://tracker.webtorrent.io:80
```

But I thought it would be better to have the options of providing this as part of the sbot infrastructure assuming folks using sbot as an application framework will probably want everything prepackaged(?). 

It the centralised aspect of the tracker is something we are not comfortable with then we could look to use a DHT implementation. 
https://github.com/feross/bittorrent-dht 

However the DHT will still have to be bootstrapped so we will need to implement something in the public sbot in this scenario.

dominictarr:
@No9 look at the blobs or gossip plugins probably.

If we can get this to fit in nicely I think this could be good.
We don't want to have and hard coded addresses in ssb - that would be re-centralize it.

I wonder If we could just use the pubs for a DHT? each pub probably knows about other pubs, and could get into the DHT that way... isolated pubs might end up with their own dht, but that could be a feature not a bug.

pfraze:
Hm, is this something ephemeral messages could handle?

No9:
@dominictarr thanks for the tips will dive in again soon,. 

RE DHT between pubs. Do we know explicitly when it's a pub that links as opposed to a client?

ahdinosaur:
> Basically a tracker keeps a list of who has got the actual content of a torrent. 

wouldn't the pubs (and any peer for that matter) be able to respond with the addresses of peers they are connected to that they know (through some message generated upon seeding a torrent) have the content?

the pubs shouldn't be too different from normal peers, just more public. i'm not sure if a DHT is necessary given ssb solves the same problem in a different way.

No9:
Hi @ahdinosaur 
You are correct the peers would be able to respond with the addresses. 

But I think we are also looking to leverage the multisource and pause/resume capability of torrents. 

i.e. downloading parts of a large piece of data from mutliple seeds as well as keeping the flexibility of allowing a peer to delete the content from there local system. 

We could look to solve the multisource part with our own implement and use the " ephemeral messages " @pfraze mentions to handle the discovery but my thinking was that torrent would get us there a lot quicker. 

That said I am happy to take suggestions if the multisource piece is not as hard as I understand it for SSB.

No9:
Ok so I have gone a little further with the DHT implementation 
https://github.com/ssbc/scuttlebot/compare/master...No9:bittorrent-dht 

I don't want to make a pull yet as this is outstanding
https://github.com/feross/bittorrent-dht/issues/47 

However following on from the suggestion from @dominictarr 
> We don't want to have and hard coded addresses in ssb - that would be re-centralize it.

I have put an event into the index.js to get hold of the ```address.host```

https://github.com/No9/scuttlebot/commit/054528f157e9d2d585ce21640c4831fd1e2be35c#diff-168726dbe96b3ce427e7fedce31bb0bcR158

The idea is that the DHT will try and connect to a potential peer on another host and thus bootstrap itself. 

Anyway I just wanted to confirm that that is the right place to raise that event from?

dominictarr:
@feross needs to be in on this discussion.

Sorry @No9 just trying to understand what is going on here - you need to know the address of another server to bootstrap your self, so this is pulling addresses off each connection and assuming that (seems reasonable) that it's another tracker?

Have you looked at the gossip plugin? that already has code for getting the available peers that are recorded in the feeds (i.e. pub servers)

dominictarr:
or are you emitting information about the client connections and connecting back to those too?

No9:
@dominictarr 
> you need to know the address of another server to bootstrap your self, so this is pulling addresses off each connection and assuming that (seems reasonable) that it's another tracker? 

That's the plan ^ 

I looked at getting the gossip plugin from the rpc object when the rpc connect/authorised events fired.
The gossip plugin is there put the peers don't seem to be populated.

Maybe I should ignore the empty ones and wait for an event when they populate?
But that didn't feel right.

feross:
@No9 The issue you reported (https://github.com/feross/bittorrent-dht/issues/47) is fixed as of bittorrent-dht 3.1.0.

dominictarr:
@No9 can you show me the code you where using? I think hooking into the gossip plugin is the right idea here

No9:
Super thanks @feross 
@dominictarr ill rehydrate where I got to with gossip and let you know 

/me wished I used git stash more

No9:
@dominictarr cleaned up my downstream which has closed this off. 
I have the remote connection coming through from the gossip plugin event. 'rpc:connect' as you suggested 
I am picking up the _remoteAddress property
https://github.com/No9/scuttlebot/blob/bittorrent-dht/plugins/torrent-dht.js#L49 
Ignoring local addresses.
Then parsing out the _remoteAddress to get the IP that I can then suffix the dht port on. 
If there is a better way to get the remoteAddress then that would be cool.

dominictarr:
During the time that this pull request has been in progress, we have decided that the best thing to do right now would be to simplify and polish ssb and get it out the door. Although there are many things that could be built on top of this pattern - and those visions where very important motivation in when we where getting started, now we have something that is working - we don't need to do that much to ship something that has not been shipped before, just details.

In that note, although this is a pretty good idea, and certainly adds value to ssb, we are already in totally new territory - and maybe we should leave this for now? I certainly appreciate the work that has gone into this, but to get this out the door we need to reduce the work that needs to be done.

Can we back burner this for now?

No9:
@dominictarr Thanks for the heads up on that and the help understanding the SSB patterns :)

I agree this isn't a 1.0 feature or possibly not even a core feature at all. 

This feature focuses on a specific use case I have for SSB and so I will keep kicking it around.

Maybe we should look at streamlining the plugin system a little more so it can be a third party plugin?
Specifically items like https://github.com/ssbc/scuttlebot/compare/master...No9:bittorrent-dht#diff-168726dbe96b3ce427e7fedce31bb0bcR338 
Where I had to update index.js to get the option accepted

PR #160, opened 2015-04-09
more convienient queries with mynosql-query

dominictarr:
this adds a special case to the cli tool to use mynqsql-query to parse queries.
it makes it easy to query the data base with a one liner.

for example, get all messages where the first message (after init) is a post

``` bash
sbot query 'sequence=1, content.type=post'
```
will produce a lot of JSON output.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -10,6 +10,7 @@ var explain    = require('explain-error')
 var ssbKeys    = require('ssb-keys')
 var stringify  = require('pull-stringify')
 var createHash = require('multiblob/util').createHash
+var parse      = require('mynosql-query')
 
 var config  = require('ssb-config')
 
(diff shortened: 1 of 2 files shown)


bin.js
```diff
@@ -60,8 +61,11 @@ function usage () {
 var opts = require('minimist')(process.argv.slice(2))
 var cmd = opts._[0]
 var arg = opts._[1]
+
 delete opts._
 
+console.error(parse(arg))
```
pfraze: is this debug?

bin.js
```diff
@@ -157,6 +161,14 @@ function next (data) {
         throw explain(err, 'auth failed')
     }
 
+    console.log(cmd, arg, data)
```
pfraze: same question

dominictarr:
fixed

pfraze:
ok cool, looks good to me

dominictarr:
okay merging

dominictarr:
merged into 4.2.0

PR #161, opened 2015-04-12
Init empty feed when starting server

clehner:
This should cause a server with an empty feed (e.g. newly created keypair) to publish an init message when it starts serving. This would address ssbc/phoenix#367 by ensuring that pub servers have profiles, and that their createdAt timestamp will be the time at which the server first started serving.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -72,6 +72,12 @@ if(cmd === 'server') {
   require('./').init(config, function (err, server) {
     if(err) throw err
 
+    // Ensure the feed is init'd
+    server.ssb.getLatest(keys.id, function (err, msg) {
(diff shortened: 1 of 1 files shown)


pfraze:
Im cool with this, @dominictarr you have any objection?

clehner:
Not ready to merge yet... I need to test this more.

dominictarr:
init messages are no longer used. maybe this is still useful though, but should this be in scuttlebot or phoenix/patchwork?

pfraze:
Patchwork has a UI flow for setting name and profile picture on first load. Maybe we do something similar for pub nodes in the CLI? Theyll need to do both.

PR #162, opened 2015-04-13
Document the new port number being used.

ootoovak:
This caught me out the first time until I got a chance to talk to @ahdinosaur and he told me about the new port. I know see this is in https://github.com/ssbc/ssb-config/blob/master/index.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ The `whoami` command outputs your id and your public key. There are many other c
 
 ### Web UI
 
-Once the scuttlebot server is running, you can access the Web UI at `http://localhost:8008`.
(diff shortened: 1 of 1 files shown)


pfraze:
8008 *is* the new port number. What happened is that the change to 8008 was accidentally half-completed. If you look in https://github.com/ssbc/phoenix/issues/306, you can see backlinks to all the open PRs that need to be merged to complete this.

Sorry about the confusion! I found out about this over the weekend but have been offline to give my head a rest. We'll finish these merges asap.

ootoovak:
All good, thanks for the background. I just wanted to get the docs inline with the code so if the code is changing to meet the docs then serves the same purpose. :+1:

PR #165, opened 2015-04-17
Fixes for local feeds

pfraze:
Outdated pull-ws-server dep was keeping the auth code from correctly detecting local auth.

Fixing that caused the invites tests to fail, as the test user was correctly authing locally, which didnt have the `invite.use` perms. Fixed that by giving `invite.use` access to the local role.


pfraze:
published `4.2.1`

PR #166, opened 2015-04-18
add perms to blobs.has and blobs.get to local role

pfraze:
fixes a bug in the blobs plugin (shown by the blobs tests) that was introduced when the local role began correctly being recognized


pfraze:
published `4.2.2`

PR #170, opened 2015-04-25
mention Node in the install

mixmix:


ahdinosaur:
nvm link is broken, maybe link to <https://github.com/creationix/nvm#node-version-manager->?

pfraze:
why not link to https://nodejs.org/ ?

mixmix:
their install gives you a tar.gz file and no help :(

you can change it to nodejs if you like. I think mentioning that it's a node project to someone who's quite new would be a kindness though

dominictarr:
yes this is good - we need to grow this beyond the node community

pfraze:
done int he docs

Issue #175, opened 2015-05-09
blob sync needs to handle a larger batch volume


pfraze:
Blob download only seems to pull one blob per connection. During new-user onboarding this takes a while, especially when lots of peers are being polled and only one or two of them have the blobs you want.

This sucks for avis. Ideally a new user would use an invite code, and pretty immediately receive the user list and see their avatars.

dominictarr:
that is certainly a bug.

dominictarr:
this is merged so closing

PR #178, opened 2015-05-11
Improve blob sync throughput and correctly release the blobs-task on completion

pfraze:
This PR addresses #175. Changes....

 - Each remote is now queried on connect independently of the other remotes (it's no longer part of a oneTrack job queue). If any blobs are found, they are queued for download. When all found are downloaded, the blob's task is released.
   - Note, only 20 blobs are queried at once, and so a single connection will download no more than 20 blobs before releasing. (We'll want to enforce this on the serving-end at some point.)
 - The download job-queue now allows up to 5 simultaneous downloads, and only waits 300ms before queueing up the next download. This increases the throughput dramatically.

### DIFF
--- a/plugins/blobs.js
+++ b/plugins/blobs.js
@@ -304,13 +304,17 @@ module.exports = {
         return done(true)
 
       // download!
+      f.state = 'downloading'
       sbot.emit('log:info', ['blobs', id, 'downloading', f.id])
       pull(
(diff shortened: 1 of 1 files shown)


pfraze:
@dominictarr good to merge?

dominictarr:
okay lets merge this, but I think we can find a better method than just limiting to 20 blobs perconnection.

Issue #176, opened 2015-05-09
prefer to sync with pubs that have a lot of unique pub-announcements


pfrazee:
I'm working on onboarding and finding that clutter in the list of announced pubs really slows down the sync process. There are 12 pubs in my table and only 3 of them are valid

The gossip plugin chooses randomly from the peer table, but weights the probability of selection. RN the only factor is number of previous failures (it's less likely to select a peer if past connect attempts have failed). I suggest we also increase the odds for pubs which have a lot of unique users publishing announcements for them.

dominictarr:
This is a great idea!

PR #181, opened 2015-05-13
Gossip with popular pubs

pfraze:
Per #176, makes gossip prefer pubs with many unique announcements

### DIFF
--- a/plugins/gossip.js
+++ b/plugins/gossip.js
@@ -52,25 +52,91 @@ module.exports = {
 
     var config = server.config
     var conf = server.config.gossip || {}
+    var host = config.host || nonPrivate.private() || 'localhost'
+    var port = config.port
+
+    // Peer Table
+    // ==========
 
     //current list of known peers.
     var peers = []
-    var seeds = config.seeds
-    seeds =
-      (isArray(seeds)  ? seeds : [seeds])
+    function getPeer(id) {
+      return u.find(peers.filter(Boolean), function (e) {
+        return e.id === id
+      })
+    }
 
+    // track connection-state in peertable
+    server.on('remote:authorized', function (rpc, authed) {
+      //don't track cli/web client connections
+      if(authed.type === 'client') return
+
+      var peer = rpc._peer
+      if(!peer) //incomming connection...
+        peer = getPeer(rpc.authorized.id)
+      if(peer) {
+        peer.id = rpc.authorized.id
+        peer.connected = true
+      }
+    })
+
+    // populate peertable with configured seeds
+    var seeds = config.seeds
+    seeds = (isArray(seeds)  ? seeds : [seeds])
     seeds.forEach(function (e) {
       if(!e) return
       var p = toAddress(e)
       if(p) peers.push(p)
     })
 
-
-    function get(id) {
-      return u.find(peers.filter(Boolean), function (e) {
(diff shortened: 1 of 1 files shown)


pfraze:
rfr

dominictarr:
looks okay - how does it change performance? is there anyway we can measure how well it works?

pfraze:
Hmm I suppose I could write a sim to measure that, and that'd be worth having to test other ideas, I'll do that

pfraze:
Ok, I put together a sim at https://github.com/pfraze/ssb-gossip-sim. Here's what this gives us:

- It generates a random graph of 100 nodes with full transitive connectivity, but with only 1/50 of the edges needed for full direct connectivity. You can see how many hops out from node 0 each node is in the sim output (below).
- It gives node 0 a "datum" and then begins gossiping, counting how many rounds of gossip it takes to get the datum from node 0 to every other node.
- A gossip round is an iteration of all nodes. If the node has the datum, it attempts to connect to a peer. If that connection succeeds, the datum is spread to the peer.
- Edges represent the ability to connect between nodes. If a node has an edge to a peer, the connection "succeeds"

With that setup, the simulator tests different ways to select the peer for a connection attempt. I put in four selection algorithms:

1. Randomly choose from all known peers (`random % 100`)
2. Randomly choose from all reachable peers (`random % num_my_edges`)
3. Randomly choose from all known peers, weighted by past failures (current alg)
4. Randomly choose from all known peers, weighted by past failures and node popularity (# of inbounds, this PR's alg)

The second method represents a sort of perfect world, where we can know which connections will fail ahead-of-time. It's there as a reference against the other methods, which are closer to reality.

Here are the results of one run:

```
Graph generated, connectivity from node #0... 
1 hops: 1
2 hops: 6
3 hops: 25
4 hops: 36
5 hops: 25
6 hops: 6
7 hops: 1

SIM RESULTS
A datum was gossiped across the network, starting from node #0, using...
Random selection from all nodes: 602 rounds
Random selection from nodes' edges: 14 rounds
Random selection from all nodes, weighted by past failures: 417 rounds
Random selection from all nodes, weighted by popularity and past failures: 119 rounds
```

It's not a perfect sim, but I think it's pretty good, and it seems to show what I'd intuitively expect. We'd get even better results if the sim was allowed to run for multiple "gossipings" of datums, since methods 3 and 4 get more effective over time.

pfraze:
PS, graphmitter is the bees knees for this kind of stuff

pfraze:
I modified the sim to disperse 100 values from node 0 instead of just 1. It would wait for a datum to gossip fully before moving to the next. The results are much more interesting:

```
Graph generated, connectivity from node #0... 
1 hops: 1
2 hops: 2
3 hops: 8
4 hops: 26
5 hops: 37
6 hops: 22
7 hops: 3
8 hops: 1

SIM RESULTS
100 datums were gossiped across the network, starting from node #0, using...
Random selection from all nodes: 25460 rounds
Random selection from nodes' edges: 1047 rounds
Random selection from all nodes, weighted by past failures: 3201 rounds
Random selection from all nodes, weighted by popularity and past failures: 2323 rounds
```

The weighted algorithms start to show a lot more value when they have time to collect failure knowledge.

dominictarr:
okay great, lets :+1:

PR #192, opened 2015-05-28
Tracking PR for replication performance

dominictarr:
Adds a crude benchmark and some fixes to make replication faster.

benchmarks of a fullish network (100 users, ~ 1000 messages) to a new user. It's currently taking 1 minute, and about 17 seconds to replicate 1000 messages. including generate 1k messages.
To replicate the receiving client verifies about 400 signatures (if it queues a few messages it just verifies the last one in the current queue).

I also benchmarked some signing libraries (not in this PR though) - eccjs (which we use) can verify 100 signatures in 4 seconds. Slow. To verify 400 signatures would take 14 seconds, which accounts for most of the time to replicate.

I also looked at some other libraries - ecma-nacl is twice as fast (in browser or server). elliptic is way faster on node but slower on the browser (? not sure why ?)

But, I think we should go with nacl. This is already a port of a specific C implementation, which will be really fast (although the node bindings are currently not working... https://www.npmjs.com/package/nacl)

NOTE: this pr breaks the test/server.js because that test doesn't detect the successful replication correctly anymore. Implementing a progress bar would solve this.

### DIFF
--- a/plugins/friends.js
+++ b/plugins/friends.js
@@ -2,6 +2,7 @@ var Graphmitter = require('graphmitter')
 var pull        = require('pull-stream')
 var mlib        = require('ssb-msgs')
 var memview     = require('level-memview')
+var pushable    = require('pull-pushable')
 
 function isFunction (f) {
   return 'function' === typeof f
@@ -15,7 +16,8 @@ exports.name = 'friends'
 exports.version = '1.0.0'
 exports.manifest = {
   all  : 'async',
-  hops : 'async'
+  hops : 'async',
+  createFriendStream: 'source',
 }
 
 exports.init = function (sbot) {
@@ -36,6 +38,7 @@ exports.init = function (sbot) {
             graphs.follow.edge(msg.value.author, link.feed, true)
           else
             graphs.follow.del(msg.value.author, link.feed)
+
         }
         if ('trust' in c) {
           var trust = c.trust|0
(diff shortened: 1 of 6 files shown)


dominictarr:
oh yeah, and there is a event leak too... probably because friend events are not being cleaned up properly.

plugins/friends.js
```diff
@@ -60,6 +63,28 @@ exports.init = function (sbot) {
...
+      var graph = graphs[opts.graph || 'follow']
+      if(!graph)
+        return pull.error(new Error('unknown graph:' + opts.graph))
+      var cancel, ps = pushable(function () {
+        cancel && cancel()
+      })
+      var conf = config.friends || {}
+      cancel = graph.traverse({
+        start: start || sbot.feed.id,
+        hops: opts.hops || conf.hops || 3,
```
pfraze: should `hops` default to 3 here? I think I'd expect it to be 1 if unspecified, because I'd be thinking "friends" would be "immediate friends"

pfraze:
Ok cool, glad you did this.

For crytpo I suggest we use a C version with node-bindings (maybe we can fix that broken one or find another working one). I think a javascript port risks introducing subtle changes which open up timing attacks or reduced entropy, for instance. (It's not that JS couldnt do it, but we have to trust the base library AND the person/process that ported to JS, and I'd rather just dodge that if we can.)

iirc What we probably want is libsodium bindings, since libnacl isnt very portable.

> oh yeah, and there is a event leak too... probably because friend events are not being cleaned up properly.

it's unreal what a pain something like that is to debug in node/js

dominictarr:
good point - though the js port looks pretty well done to be honest, it would be much better if we can test it to be exactly compatible with a binding.

I've made a quick hack to ssb-keys to support ed25519 via ecma-nacl and the test is replicating twice as fast!!!

pfraze:
niceee

dominictarr:
okay this is way faster now if we upgrade the deps to use the latest ssb-keys, i was able to increase the size of the test 100 times and it's still faster.

I did run into something weird when it asks for ~ 350 feeds at once. Don't know what this is yet, feels like leveldb though.

pfraze:
:boom:

dominictarr:
okay this is passing the tests now.

pfraze:
@dominictarr could you explain what the aborter in the replication plugin does?

dominictarr:
sure--that kills the stream that listens to the follow graph. it's needed because that stream isn't a part of the replication steam, it just sinks to a drain that sets up the individual streams (it's basically a stream of stream) https://github.com/dominictarr/scuttlebot/blob/replication_performance/plugins/replicate.js#L17-L32

We need to abort that when the actual replication stream ends, so that the listener on the follow graph is cleared. https://github.com/dominictarr/scuttlebot/blob/replication_performance/plugins/replicate.js#L45

pfraze:
Gotcha, thanks

dominictarr:
hmm, seems this stalls when running on travis...

PR #195, opened 2015-06-15
Blob prioritization by requests, age, and failure rates

pfraze:
Prioritizing blobs by recency, user interest, and failure-to-fetch

### DIFF
--- a/plugins/blobs.js
+++ b/plugins/blobs.js
@@ -149,7 +149,6 @@ module.exports = {
 
         if (cb)
           wL.byId[hash].waiting.push(cb)
-        wL.byId[hash].requests++ // track # of requests for prioritization
 
         if (isnew) {
           // trigger a query round with the existing connections
@@ -367,12 +366,16 @@ module.exports = {
 
         blobs.has(hash, function (_, has) {
           if (has) return cb(null, true)
+          
+          // update queue
           if (nowait) {
-            wantList.queue(hash)
-            cb(null, false)
+            wantList.queue(hash); cb(null, false)
           } else {
             wantList.queue(hash, cb)
           }
+
+          // track # of requests for prioritization
+          wantList.byId[hash].requests++
         })
       },
 
--- a/plugins/blobs.js
+++ b/plugins/blobs.js
@@ -1,5 +1,8 @@
 'use strict'
 
+//                 ms    s    m    h    d
+var MONTH_IN_MS = 1000 * 60 * 60 * 24 * 30
+
 var Blobs = require('multiblob')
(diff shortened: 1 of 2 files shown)


pfraze:
The prioritization increases linearly by # of requests, decreases linearly by # of not founds.

The `requests` counter is increased by 1 every time the blob appears in a feed message, or when `want()` is called for the blob. The `notfounds` counter is increased by 1 every time a `has()` query to a remote comes back `false`.

The `want()` function now has a `{nowait:}` option so that an application can increase the req count without queueing a long-term CB. When `nowait: true` is set, the behavior is like `has()` plus the increased `request` count:

```js
sbot.blobs.want(hash, {nowait: true}, function (err, has) {
  if (!has) {
    console.log('dont have it currently, but increased priority')
  }
})
sbot.blobs.want(hash, function (err, has) {
  console.log('blob has arrived')
})
```

Todo
 - A user could publish the same blob link N times to increase its prioritization. We could track the author of a blob link, and only let each feed increment the priority once, but I don't think that's worth the memory or diskspace usage to track. Instead, I think the feeds should only queue the blob, and never increment its `requests` counter.
 - Attempts to access a blob from phoenix's UI should increase its priority, but, for something very common like a profile image that can't be found, this could lead to runaway growth on the `requests` counter. I'm going to cap the `requests` counter at 20 to mitigate this. Likewise, I'm going to cap the `notfounds` counter at 40. The final priority will be calculated as `clamp(requests - notfounds, -20, 20)`. This means a highly-requested blob's priority can range from 20 (`20 requests - 0 notfounds`) to -20 (`20 requests - 40 notfounds`), while an infrequently-requested blob will range from 0 to -20. This, with a small random modifier on the priority (1-5 points), will keep the queue from always requesting the same set of popular but failing blobs.
 - Blobs which were published a long time ago, but which haven't been requested by the user, may not be worth searching for. I'm going to only auto-queue blobs from the feed if the timestamp is plus/minus a month from the current date. If the user does request the blob, the UI will call `want()` and put it into the search queue.

pfraze:
@dominictarr this is ready for review
I'm going to do rate-limiting in another PR because that may relate to other parts of sbot

dominictarr:
instead of a double-negative it could be `wait:false` or `asap: true`.

I'm suspucious of strategies which have fixed absolute values, but this is better than what we have now, so lets merge!

pfraze:
yeah, I dont like using false for flags that default to a falsey value. `asap` seems imprecise, maybe `immediate`?

I know what you mean about the absolute values, we'll see how it goes

PR #196, opened 2015-06-16
blob RPC usage limits

pfraze:
This PR adds resource-usage tracking to connections, and removes permissions to the RPC API when configured thresholds are exceeded. Connections authorized as 'master' won't be subject to limiting.

This will track only resource usage in a single connection's life time. It will not handle multiple connections by a given user, or the server's overall resource availability.

This PR will track and limit the usage of:

 - blobs: # of files queried for
 - blobs: # of files transferred

In the future, we may also want to track and limit:

 - blobs: # of bytes transferred
 - # of RPC calls in the connection
 - # of RPC calls within a window (no flooding)
 - # of feeds replicated
 - connection lifespan (eg cant hold open connection > 5 mins)

Tracking will be handled by attaching a `usage` object to the `rpc` object. RPC methods will increment relevant metrics and check against configured limits before running their behaviors. In the future, we may want to bake the tracking and limiting into muxrpc, but I decided to have functions do it "manually" while we learn the problem space.

Config for limiting is under the `limits` object. The `ssb-config` repo will need updating. The additional config is:

```js
{
//...
  limits: {
    blob: {
      queries: 5, // I have this low because it's usually done in bulk
      transfers: 50
    }
  }
//...
}

### DIFF
--- a/index.js
+++ b/index.js
@@ -90,6 +90,13 @@ exports = module.exports = function (config, ssb, feed) {
 
     rpc.incoming = incoming
     rpc.outgoing = !incoming
+    rpc.usage = {
+      blobs: {
+        queries: 0,
+        transfers: 0,
+        bytes: 0
+      }
+    }
 
     rpc._sessid = ++sessCounter
     rpc._remoteAddress = stream.remoteAddress
--- a/plugins/blobs.js
+++ b/plugins/blobs.js
@@ -49,6 +49,16 @@ function clamp (n, lo, hi) {
   return Math.min(Math.max(n, lo), hi)
 }
 
+function opt (obj, path, fallback) {
+  path = path.split('.')
+  while (path.length) {
+    obj = obj[path.shift()]
+    if (!obj)
+      return fallback
+  }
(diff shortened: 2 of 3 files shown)


dominictarr:
revoking permissions isn't the right way to handle this... you should just drop the connection

dominictarr:
if there isn't an api to do that we can add one, because there is already a thing to kill connections that go quiet... that could be adapted to have either a maximum rate, or a kill switch or both

pfraze:
"remove permissions" I wrote that before implementing. It just returns an error.

Dropping the connection would be extreme if it's not an attack, if it interrupts any tasks in progress. Id expect a blob client to try to request blobs as long as it's allowed, then stop trying when it gets the rate-limit error. We need a way to the client to discern that, muxrpc doesnt have error codes

dominictarr:
moderating resource usage has turned out to be quite tricky, this one has a pretty obvious work around, just reconnect - am gonna close this we'll probably need to do some more serious thinking about how this needs to work.

PR #197, opened 2015-06-17
tracking PR to add encrypted channels

dominictarr:
This is a tracking pull request to add point-to-point encryption via secret-handshake.
>note, this pull request is made on top of the replication perf PR.

so far I have test/server.js passing.

There is quite a simplification that can happen here, for example, since secret-handshake
handles mutual authentication we can remove all the auth stuff.


dominictarr:
okay this also rewrites the cap code. special cases like hmac auth are gone.
to mint a cap, generate a private key from a seed, and share your `ip:port,publicKey,seed`
saving permissions you will give them.
then using cap-client a peer may login (fixed number of times) and use the api you have restricted them to.

TODO: refactor the create cap system out of the invite plugin and make it generic.

dominictarr:
this is now passing all the tests at my end.

dominictarr:
todo for after this
* rethink browser logon, possibly with secret-handshake working in the browser?
* kill emit permission. it's used to track when tasks are done, and in blobs.
* better muxrpc encoding... so that buffers do not need to be base64 encoded.

pfraze:
Looking great so far! Got some questions

 1. What's the purpose of `client.js` versus `ssb-client`, do we just need to merge those?
 2. What was the method for producing the ssb protocol cap?

dominictarr:
@pfraze I added a comment explaining it: https://github.com/dominictarr/scuttlebot/commit/24e86c06f8660a0ce91f477d7c892e9f6cd846ba

pfraze:
cool, what about client.js?

dominictarr:
oh yeah, that functionality should be moved to ssb-client

dominictarr:
Okay this will currently not be working with phoenix, but I think we have enough here to merge a major version.
I'll work on ssb-client next.

dominictarr:
oops, I still need to rewrite invites! no merge yet.

dominictarr:
Okay, turns out invites was easy! Aliased `invite.accept` to `invite.addMe` I don't know why I didn't think of that sooner...

dominictarr:
removed `done` event... and tests still passed. blobs are the only events left now

dominictarr:
(rpc events, i mean)

dominictarr:
permissions for the emit method are now gone. instead i wrote https://github.com/dominictarr/pull-notify 
allows an event emitter of pull streams.
(see also: https://github.com/ssbc/muxrpc/issues/23)

pfraze:
:+1: :confetti_ball: lots of great stuff in this one

PR #199, opened 2015-06-26
switch to tcp and delete everything I could

dominictarr:
this is on top of https://github.com/ssbc/scuttlebot/pull/197

it switches to tcp instead of ws, and deletes all the code I could.


dominictarr:
I made this pr instead: https://github.com/dominictarr/scuttlebot/pull/3

PR #200, opened 2015-06-29
Blobsmerge

dominictarr:
I added my changes, but something is holding the test blob test you added open.
(I moved it into `test/blobs3.js`, since the blob tests had a tendency to take ages at it was easier if you could run them separately)


pfraze:
@dominictarr just a heads up, to `npm install secret-handshake` I had to `brew install automake libtool`

dominictarr:
@pfraze oh, I must have had that already... we can set up prebuilt binaries like leveldown uses now.

dominictarr:
then it will work easily on windows etc

pfraze:
Aha, found the bug: no `key` on the seed address for replication

PR #201, opened 2015-07-01
replace trust-graph with flag-graph in friends

pfraze:
This is to support the change to a `flag` attribute in `type: contact` messages.

### DIFF
--- a/plugins/friends.js
+++ b/plugins/friends.js
@@ -24,7 +24,7 @@ exports.init = function (sbot) {
 
   var graphs = {
     follow: new Graphmitter(),
-    trust: new Graphmitter()
+    flag: new Graphmitter()
   }
   var config = sbot.config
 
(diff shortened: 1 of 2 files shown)


dominictarr:
flag graph only has negative edges though right?
I don't care that someone I flagged flagged someone.
I flagged Bob because he is an asshole. He flagged all my friends because he is an asshole.

But, I care about who my friends flagged... that was what the positive trust edges stood for.

dominictarr:
I think what we need is some thing that "approximates" trust, that "trust" can be infured from... (prehaps membership is a friends or family group?)

Probably "friends" because you can stop being friends with someone, but you can't stop being family

pfraze:
I can rewrite phoenix not to require the `flag` graph -- phoenix-api's profiles track the same info. In fact, I'm going to do that to cut down on RPC calls.

RE trust, We currently have the "follow" signal. You only see the flags and follows by people you follow. It approximates, "I trust this person enough to see what they have to say." I think that's good enough for now. We don't take name- or picture-assignments from other users.

dominictarr:
yeah, I think "follow" is a suitable "trust" proxy for now

pfraze:
We can merge or close this if we want, or replace it with a PR that removes the `flag` graph. Phoenix no longer depends on either function.

dominictarr:
okay lets remove this

pfraze:
close PR or remove `flags` graph?

dominictarr:
remove flags

dominictarr:
by the way, i merged this into by blocked pr.

PR #202, opened 2015-07-01
use phoenix-api plugin

pfraze:
This is needed to make phoenix work


dominictarr:
I thought you where gonna require sbot from electron-scuttlebot now, so you should be able to call .use from there?

pfraze:
Oh ok, that's a good idea.

PR #203, opened 2015-07-02
Updates to support a sync-progress UI in phoenix

pfraze:
These changes will make it possible to implement the "initial sync" ui, and a general network-sync UI, in phoenix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "osenv": "~0.1.0",
     "pull-paramap": "~1.1.1",
     "rimraf": "~2.2.8",
-    "ssb-msg-schemas": "~3.1.0",
+    "ssb-msg-schemas": "~3.2.0",
     "tape": "~4.0.0"
   },
   "bin": {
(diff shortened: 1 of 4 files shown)


pfraze:
@dominictarr it would be awesome if we could show a progress bar with % complete in phoenix, especially for the initial sync.

We'd need to know how many messages we're downloading. To get that we either need to modify `createHistoryStream` to send that data, or we need to add an extra RPC call to the remote. Could do the extra call in parallel - if the response doesnt come before replication ends, no biggy.

dominictarr:
@pfraze actually we do know that because _they_ send their `createHistoryStream` so you know what they are up to, we just have to intercept the incoming `createHistoryStream` arguments.

dominictarr:
hmm, i started working on getting progress out of a replication stream... and I found something very weird.
If I gave pull-paramap a width parameter (https://github.com/ssbc/secure-scuttlebutt/blob/master/index.js#L326-L331) (asking it to only make that many in flight requests at once...) then things worked, else they went weird. If there is no width parameter, paramap makes requests as fast as it can.

This gives me a hunch about maybe somethings sometimes took too long... too many concurrent requests to the database? there are several other uses of paramap without widths (maybe solution is to set a default width?)

dominictarr:
:+1:

pfraze:
published `5.2.0`

PR #204, opened 2015-07-02
Replication Progress

dominictarr:
This allows you to track real time replication progress.
when you run `test/random.js` you'll see a simple ascii progress bar, too.

The replication plugin adds a progress observ(able) to the rpc object,
and if you pass it a listener it'll be updated with the current progress.
It is only updated every 100 ms, max.


pfraze:
:+1:

pfraze:
travis is getting hitched up during npm install for some reason, but one of the tests passed, going to go ahead and merge

PR #205, opened 2015-07-03
use key instead of hash(key) as rpc.id

pfraze:
Using `hash(address.key)` was causing a bug in `invite.accept()`, as it was following the `rpc.id` and therefore not following the feed id


pfraze:
@dominictarr any problems with this?

pfraze:
@dominictarr test failed in one travis run but not the other. Race?

pfraze:
Looks like there was a bug in test/random, should be fixed now

dominictarr:
merged into 5.1.0

PR #208, opened 2015-07-06
get everything working with Sha256

dominictarr:
okay! this is it!

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,12 +18,13 @@
     "map-merge": "~1.1.0",
     "minimist": "~1.1.0",
     "mkdirp": "~0.5.0",
-    "multiblob": "~1.4.3",
+    "multiblob": "^1.5.0",
     "multicb": "^1.0.0",
     "muxrpc": "~5.0.1",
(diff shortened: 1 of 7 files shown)


pfraze:
:+1:

ralphtheninja:
Why are you switching?

dominictarr:
@ralphtheninja changing all the crypto to C implementations, for speed.

This made a tremendus difference with signatures, since asymetric crypto is in general much more expensive that symmetric*, but switching to JS hashes made generating a large dataset (https://github.com/ssbc/ssb-testnet) 30% faster. Sodium does have a blake2b,
but i needed to make a PR to the bindings, and it wouldn't be compatible with blake2s anyway.

(*here i did some simple perf tests learn the relative costs of the various primitives: https://github.com/dominictarr/sodiumperf)

dominictarr:
anything needed before we merge?

pfraze:
dont think so, Ive been working with this branch for a bit

dominictarr:
am merging this.

PR #211, opened 2015-07-14
Tracking: implement blocking.

dominictarr:
have also refactored auth api to simplify integration.
currently blocked clients are prevented from connecting (to the blocker)
next is update to replication plugin to prevent blocked requesting blocker's data.


pfraze:
:+1: looks good

dominictarr:
still a bit more work to do here. tests are failing because this depends on the latest secure-scuttlebutt but we havn't merged that pr yet

dominictarr:
this is now fully functional, but just needs to be refactored so that it's not so tightly coupled.

dominictarr:
refactoring is taking a bit longer than i intendend, so i think we should just merge this, and then it won't be stalling anything.

pfraze:
Go ahead if you want to, but Im actually blocked until the link changes land, so it doesnt matter too much whether this PR merges. (Not 100% blocked, fwiw.)

dominictarr:
which link changes?

pfraze:
The sigils and message structure changes

dominictarr:
right, okay cool

PR #212, opened 2015-07-27
Refactor

dominictarr:
This refactors out the core elements into secret-stack (formerly "illuminati")
This makes the core ssb much smaller, and also, the plugins are now much more self-contained.


dominictarr:
the way that instances of sbot are instantiated has changed, so it will probably be necessary to update anything that uses this.

dominictarr:
this now integrates all the sigil changes.
ready to merge.

dominictarr:
okay so travis is being weird... so far i have figured out that it's taking an ipv6 address on travis,
but i've had trouble reproducing that on my machine. I can't seem to get ipv6 addresses to work as i'd expect on my machine at all.

dominictarr:
merged into 6.0.0

PR #213, opened 2015-08-03
Readd index rebuilding

pfraze:
Conditional index-rebuilding got lost in the refactor

This is where, on load, SSB compares a vmajor stored in level against the vmajor of the SSB package.json. If they aren't equal, the type and link indexes are deleted, and the log is replayed to create new indexes.

The secret-stack init process doesn't support async work, which this unfortunately requires. To solve that, I exposed SSB via a `getSSB()` method on the sbot object. `bin.js` then uses that to run the rebuild, as does patchwork.

Previously, rebuild happened before plugins were added. Now, it happens after. This could create problems for any plugin which relies on the type or link indexes during init. Blobs and gossip both do.
@dominictarr how should we solve this?

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -79,7 +79,19 @@ cmd = aliases[cmd] || cmd
 
 if(cmd === 'server') {
   config.keys = keys
-  createSbot(config)
+  var sbot = createSbot(config)
+  sbot.getSSB().needsRebuild(function (err, b) {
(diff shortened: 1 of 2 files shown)


pfraze:
We decided not to include this for now, since all past DBs will be invalidated by this update

PR #210, opened 2015-07-09
add publishBoxed, box, unbox to rpc api

pfraze:
Adds:

`publishBoxed(data, recipientPks, cb)`
`box(data, recipientPks, cb)`
`unbox(ciphertext, cb)`

PR #214, opened 2015-08-03
Refactor fixes

pfraze:
Readds box, unbox, and boxPublished -- those never got merged https://github.com/ssbc/scuttlebot/pull/210

adds createUserStream and whoami to the sbot apis, Patchwork depends on both

fixes the path passed to ssb

adds sbot.keys, which patchwork-api needs during processing to decrypt messages


pfraze:
moved the crypto functions into a new 'crypto' plugin

pfraze:
added `createFeed` to sbot, needed it for the phoenix-api tests, but I figure it's generally useful

dominictarr:
merged into 6.1.0

PR #215, opened 2015-08-04
tidy/refactor: clean up gossip controls

dominictarr:
this is a refactor of the gossip plugin. code was deduplicated, and concerns separated.
for example, there was `connect` which chose a peer to connect to, and then called `connectTo` which actually connected to it. refactored so to have a choosePeer function and then a connect function.

I want to take this towards separating the strict parts which can be easily tested (adding a peer, etc)
from the fuzzy parts (choosing which peer to connect to, etc)

I think we do a similar thing with blobs.

### DIFF
--- a/plugins/gossip.js
+++ b/plugins/gossip.js
@@ -5,35 +5,15 @@ var toAddress = require('../lib/util').toAddress
 var nonPrivate = require('non-private-ip')
 var u = require('../lib/util')
 
-function all(stream, cb) {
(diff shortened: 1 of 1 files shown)


dominictarr:
oh, and I also added a few features. `add` to add a peer via the api (cli for example)

pfraze:
:+1:

dominictarr:
merged into 6.1.0

PR #216, opened 2015-08-06
rimraf needed for global install

heavyk:
cheers


dominictarr:
thanks, merging.

dominictarr:
merged into 6.1.2

PR #218, opened 2015-08-06
fix: replicate plugin now returns api during init, as intended

pfraze:
quick simple bugfix


dominictarr:
oops, thank you

pfraze:
travis seemed to fail installing level

dominictarr:
merged into 6.1.2

PR #219, opened 2015-08-07
ensure that addresses are valid

dominictarr:
this change ensures that addresses added to the peer table are valid.


pfraze:
:+1:

dominictarr:
merged into 6.1.3

PR #220, opened 2015-08-08
various bug fix, and improve progress bars

dominictarr:
made these changes to get it working with patchwork.
the most important change was to progress bars.

now a progress message looks like this:

``` js
{type: 'progress', peerid: <id>, progress: p, total: t, feeds: f, sync: bool}
```
where `feeds` is the number of mutual feeds you are replicating, and `sync` is whether you are already in sync.

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,10 +42,10 @@ var SSB = {
     'whoami'          : 'async',
     'auth'            : 'async',
     'relatedMessages' : 'async',
-
     //local nodes
     'getLocal'    : 'async',
 
+    'latest'                 : 'source',
     'query'                  : 'source',
     'createFeedStream'       : 'source',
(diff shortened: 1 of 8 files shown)


pfraze:
Does the `sync` attribute mean sbot is using continuous replication?

dominictarr:
@pfraze yes. that is on by default. though, currently there is a timeout that disconnects the replication steam. I'm looking for better ways to do this though. it's worked like that for a while now.

dominictarr:
since the last time i rewrote replication, a few months back.

pfraze:
ok cool, pr looks good

dominictarr:
merged into 7.0.0

technically this is a major change since i removed the rpc.progress observable, and now there is only replicate.changes

PR #221, opened 2015-08-11
update peer-state and log when gossip connections fail

pfraze:
This was causing peers in the sync page to appear to be connecting, when, in reality, the connection attempt had failed

### DIFF
--- a/plugins/gossip.js
+++ b/plugins/gossip.js
@@ -227,6 +227,7 @@ module.exports = {
       server.connect(p, function (err, rpc) {
         if (err) {
           p.connected = false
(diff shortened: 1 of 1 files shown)


pfraze:
@dominictarr not sure what's going on with travis, I dont get any errors when I run tests at home

dominictarr:
:+1: it's something that doesn't build with iojs, but works with node.

pfraze:
ok weird, mergin

pfraze:
publish `7.0.2`

dominictarr:
to clarify the iojs thing is about building, not about running. maybe iojs has changed?

PR #222, opened 2015-08-12
Progress3

dominictarr:
this should ensure that a progress event is emited once we are in sync.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -20,6 +20,7 @@ var createSbot   = require('./')
   .use(require('./plugins/blobs'))
   .use(require('./plugins/invite'))
   .use(require('./plugins/block'))
+  .use(require('./plugins/local'))
   .use(require('./plugins/logging'))
   //TODO fix plugins/local
 
--- a/package.json
+++ b/package.json
@@ -60,9 +60,10 @@
     "tape": "~4.0.0"
   },
   "bin": {
-    "sbot": "./bin.js"
+    "sbot": "./sbot.js"
   },
   "scripts": {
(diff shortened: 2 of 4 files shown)


pfraze:
:+1:

pfraze:
@dominictarr Can you update the progress events to include the hostname and port of the peer? Key is not 100% uniquely identifiable

dominictarr:
@pfraze it ought to be. when you receive connections it will come from a random outgoing port, so it won't really be useful. I suggest we just use the key.

pfraze:
@dominictarr This condition is currently visible, your pub has an announcement for localhost and for its public id:

```
176.58.117.63:8008:@J+0DGLgRn8H5tVLCcRUfN7NfUcTGEZKqML3krEOJjDY=.ed25519
localhost:8008:@J+0DGLgRn8H5tVLCcRUfN7NfUcTGEZKqML3krEOJjDY=.ed25519
```

dominictarr:
@pfraze yes. people can publish invalid pubs, so we should filter that out. better: just use the key, and swap that for a name if available.

pfraze:
@dominictarr It's hard to evaluate which pub declarations are invalid. What if a pub is announced under two public IPs? Do we filter out based on failed connects? What if the net was just down, or what if the server is moving and the new address was announced before the move?

dominictarr:
@pfraze yes, you have to try and if it fails remove it. I guess you could flag pubs as well. but they would have to be presented in the UI in some way.

pfraze:
Yeah I think this will need to be a UI concern

pfraze:
published `7.2.8`

PR #225, opened 2015-08-22
fix some cli bugs + local peer detection

dominictarr:
just fixing a few small bugs in the CLI
and adding local peer detection, via udp multicast


pfraze:
:+1:

PR #226, opened 2015-08-23
set peer.connected to false on closed event

pfraze:
This fixes the network sync page error I was seeing


dominictarr:
merged into 7.1.2

PR #227, opened 2015-08-24
gossip.connect() now uses the peer's persistent entry, so that manual connects update peer state

pfraze:


dominictarr:
okay good idea.

pfraze:
good to merge?

pfraze:
published `7.2.7`

PR #229, opened 2015-08-26
bring setup instructions up to date

dominictarr:
--host is no longer necessary because sbot can test if it has a public ip address.
Also fixed output that hadn't been changed since we adopted sigils, and removed the bit about `master` but that needs to come back,
but needs to be simplified now that we have shs.


pfraze:
Closing, got all of this in the docs repo

PR #234, opened 2015-09-13
Tracking PR: command-line rewrite to use muxrpcli

pfraze:
Two goals: 1. factor out `muxrpcli`, and 2. provide better live help


dominictarr:
this is good but (as discussed on irc) there is some room for improvement around the usage convention.
we discussed yaml and (a strict) markdown format

pfraze:
@dominictarr ok what do you think of the `usage()` pattern now?

Things still todo:

 - need to update plugins to work with this
 - do we want to read the markdown files using `fs` at runtime? would we prefer to use `require()` so that noderify can put the markdown in the js bundle?
 - there are some weirdities to handle in the manifest: getAddress, getLocal, maybe others, were in the manifest (in `./index.js`) but not exported
 - we need docs for auth and query, @dominictarr I'm passing those to you

pfraze:
that covers all the docs except `query` and `auth`

ahdinosaur:
:+1: woot @pfraze :smile: (will try to expand on my use case once i have capacity, cheers for this though)

pfraze:
@ahdinosaur you bet! Dont know if you saw, but this will solve our CLI help issue too (via https://github.com/ssbc/mdmanifest)

ahdinosaur:
yes, really happy with the approach you've taken, separating out the cli interface (which points to the possibility of more interface modules: http, websockets, zmq, whatever) and using markdown as both docs and manifests.

pfraze:
:+1: :rainbow:

pfraze:
question: would it be better to put all api docs in one file, and then share that (and its output manifest) with the plugins?

ahdinosaur:
`¯\_(ツ)_/¯`

splitting it up seems reasonable as well.

dominictarr:
i guess calling it "plugins" implies that we are at least hoping to refactor them out at some point, so leaving them separate is probably better!

pfraze:
> there are some weirdities to handle in the manifest: getAddress, getLocal, maybe others, were in the manifest (in ./index.js) but not exported

Tests pass without, I dont know of any code that depends on them. Going to assume that was cruft that needed removing

pfraze:
Going to bump patch because this only adds docs and input validation, it doesn't add behaviors.

pfraze:
published `7.2.3`

PR #237, opened 2015-09-15
add validation

pfraze:
Wanted to get better error output for the command-line, so I'm adding this validation layer


index.js
```diff
@@ -8,6 +8,8 @@ var mkdirp     = require('mkdirp')
 var rimraf     = require('rimraf')
 var mdm        = require('mdmanifest')
 var fs         = require('fs')
+var valid      = require('muxrpc-validation')
+require('./validators')
```
dominictarr: Why are you requiring ./validators without using a return value or passing it anything? seeing a line like this makes me think that it adds globals, but it doesn't seem to do this, so it must mutate some other shared state? It's function would be way more obvious if it operated on something that was passed to it. (or did not alter state, but just returned a value)

pfraze: Ah yeah I figured you wouldnt like that. Pushed a fix

index.js
```diff
@@ -8,6 +8,8 @@ var mkdirp     = require('mkdirp')
 var rimraf     = require('rimraf')
 var mdm        = require('mdmanifest')
 var fs         = require('fs')
+var valid      = require('muxrpc-validation')
+valid.set(require('./validators'))
```
dominictarr: this is still global for `muxrpc-validation` because it mutates data in that module. this could cause problems if there is more than one muxrpc-validation dep in the tree (which one are you referencing?)
what would be better would be to have mrv return a setup function.
`var valid = require('muxrpc-validation')(require('./validators'))`

Although, at that point, you should just do:
``` js
//validators.js
module.exports = require('muxrpc-validation')({
...
})
```
and in index.js
``` js
var valid = require('./validators')
```
then the validators file entirely sets up and describes the validator functions.

pfraze: good call, done

pfraze:
Merging into cli, we can continue discussion in that branch on anything in here

Issue #245, opened 2015-09-24
resync on wake up after laptop has been suspended.


dominictarr:
This is a pretty low hanging fruit. 
Just have a setInterval that checks how much time has elapsed and if it was more than one interval, that means you have been asleep. then resync with all the pubs with high priority.

pfraze:
done

PR #247, opened 2015-09-25
trigger gossip syncs when the device wakes up from sleep

pfraze:
#245

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "non-private-ip": "~1.1.0",
     "observ": "~0.2.0",
     "observ-debounce": "^1.1.1",
+    "on-wakeup": "^1.0.0",
     "osenv": "~0.1.0",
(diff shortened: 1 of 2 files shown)


pfraze:
@dominictarr ready for review

dominictarr:
It's only ten lines, but what do you think about putting this in a separate module?
it would be a short module, but signs that it would work:

* it would do what it says on the tin (called "on-wakeup" or something)
* it's internal workings arn't relavant to anything else in scuttlebot
* it's interface is simple (takes a callback + timeout)

Also, you can use `interval.unref()` and that interval won't keep the process alive, so you don't need to close that (node only)

pfraze:
Ah, sure

pfraze:
@dominictarr ok done!

dominictarr:
looks great!

dominictarr:
merged into 7.3.3

PR #248, opened 2015-09-26
fix: decrease count of open connections in gossip on failed connect

### LINKED ISSUE #242: Avoid discouraging the gossip process
I've been noticing, after a period with an internet connection, the gossip won't resume. Looking at the sync page in PW, it's clear sync attempts aren't occurring (because the timestamp of the last attempt is old). I believe this is because each peer racks up a high failure-count during the downtime, and `plugins/gossip.js#choosePeer` doesnt require that a peer is selected.

One possible solution is to choose a peer at random, without any weighting, if the first weighted selection fails to return a hit.

pfraze:
fixes #242

### DIFF
--- a/plugins/gossip.js
+++ b/plugins/gossip.js
@@ -230,6 +230,7 @@ module.exports = {
 
       server.connect(p, function (err, rpc) {
         if (err) {
+          count = Math.max(count - 1, 0)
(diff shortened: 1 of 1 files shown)


pfraze:
published `7.3.1`

PR #258, opened 2015-10-29
replication back pressure

dominictarr:
This applies a simple algorithm to ensuring that peers do not publish too much at once.
peers accept anything from someone they follow, but only 100 items (default) per day from indirect peers.

this depends on some small breaking changes to secure-scuttlebutt

### DIFF
--- a/api.md
+++ b/api.md
@@ -287,7 +287,6 @@ getAddress(cb)
 
 
 
-
 ## getLatest: async
 
 Get the latest message in the database by the given feedid.
@@ -313,6 +312,22 @@ latest
 ```js
 latest()
 ```
+
+
+
+## latestSequence: async
+
+Get the sequence and local timestamp of the last received message from
+a given `feedId`.
+
+```bash
+latestSequence {feedId}
+```
+
+```js
+latest({feedId})
+```
+
 
 
 ## whoami: sync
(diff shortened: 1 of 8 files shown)


pfraze:
1. 1. is `latestSeqence` the same as `latest`? They share an entry [here](https://github.com/ssbc/scuttlebot/pull/258/files#diff-00e71df22262087fd8ad820708997657R318), but I'm thinking that may be a typo in the `js` example.
2. no more progress events? https://github.com/ssbc/scuttlebot/pull/258/files#diff-e736a1286f75ad71a54322d6fab5df38L48

otherwise :+1:

dominictarr:
`latest` returns a stream but `latestSequence` is just async. maybe there should be a convention about this?

pfraze:
Sometimes streams use the `create*Stream` convention. Very wordy, but it's established.

dominictarr:
yeah. maybe we can drop "create" from that though.

pfraze:
Im fine with that as a convention, would you want to update all methods?

dominictarr:
yeah, but lets alias back to what we have now. updating all that code isn't really productive. we can do that as we go.

pfraze:
:+1:

PR #261, opened 2015-11-13
tracking: quota based blob replication

dominictarr:
This change calculates blob disk usage per feed. If multiple feeds link to a blob, it's size is split between all the linkers. Then, there is a very crude cutoff. if you don't follow that feed they are allowed 20mb. I certainly think we could improve on that, but this gets us closer.

I also refactored blob replication so that it's based around an explicit queue structure. There is a separate queue per job (check if peers have something, and then get that thing) this means each of those jobs can be managed separately (has is quite light work, but getting is heavy). jobs move between queues (normally `has -> get`, but if `get` was an error, then back to `has`)

And it's easier to see what is going on. queue.js could probably be it's own module, but what it does isn't interesting enough yet.


dominictarr:
The pritorization is crude, but will be easy to adapt to say, urgently downloading something shown on the screen or that the user clicks on. (just have a function that sorts the queues).

pfraze:
Looking great

1. There's some small dev hangovers in there (https://github.com/ssbc/scuttlebot/pull/261/files#diff-a6955ec835b5f038ac275e733a64a80bR95, https://github.com/ssbc/scuttlebot/pull/261/files#diff-a6955ec835b5f038ac275e733a64a80bR49) you might just take a pass over it to get those guys
2. What's the semantic of multiblob's `ls()` `long` parameter? I see what it does, but I just dont understand why it's called `long`
3. https://github.com/ssbc/scuttlebot/pull/261/files#diff-8ab6833a6657270fadd6fbf18322c097R12 does the quota-sharing work if it's not handling new msgs?
4. Now that we're adding quotas I think we're going to need a system for garbage collection. We want randoms to stop giving us blobs at 20 MB, but we also want a new profile picture to replace their old one. If we dont get a way to clear out the old for the new, their pic will just be stuck. I think we'll need a management toolset for this. It shares properties with memory-management in language runtimes, so we might draw some inspiration there.

dominictarr:
`long` is the same as `ls -l`. see `man ls`. Better suggestions are welcome. we have other cases in ssb where there is a `meta` option (links) that adds in metadata. We could use that.

Fixed the dev hangover stuff.

The only thing left to do here is decide how to represent the quotas on the configuration.

@pfraze I think we should avoid implementing GC just yet. It's not really a problem yet - we arn't implementing quotas to save disk, but to prevent abuse.

If we have friend selected avatars that will fix the problem anyway, because your new profile will probably be on a quota of someone that is followed.

pfraze:
> long is the same as ls -l. see man ls. Better suggestions are welcome. we have other cases in ssb where there is a meta option (links) that adds in metadata. We could use that.

Oh, that makes sense. Nah that's fine.

> The only thing left to do here is decide how to represent the quotas on the configuration.

I think my question 3 needs to be addressed still, not sure the quota-sharing btwn users is effective in the form I saw it.

> @pfraze I think we should avoid implementing GC just yet. It's not really a problem yet - we arn't implementing quotas to save disk, but to prevent abuse.

It can wait, but it should stay on our radar. It's a key feature to using SSB as media-distribution channel.

> If we have friend selected avatars that will fix the problem anyway, because your new profile will probably be on a quota of someone that is followed.

The major point there: we dont know how the system will work yet.

dominictarr:
okay sure - well in any case, the total blobs disk usage I have is 70 mb. the full quota will probably be 10 - 100 times that.

dominictarr:
(I mean, for all users that you are replicating)

pfraze:
Ok we still gotta address my question 3 though, is the quota sharing effective?

dominictarr:
The quota sharing just means that if N people link to the same blob, it' size/N is subtracked from each of their quotas. However, nearly all cases currently, there is only one, and a few times 2, linkers.

dominictarr:
We could also share quota with people who have +1 a post with that link, or replied, etc.

pfraze:
Yeah I get the intended effect, but the comment I linked to in my question confirms what I'm saying -- that the calculation isn't done correct in realtime; you have to restart sbot for quota-sharing to calculate correctly.

dominictarr:
yeah. that recalculation is only gonna change when more people link to your post, reducing your quota. so it's inaccurate. but probably not much. I was about to use a thing that restarted the calculation periodically, but decided this was much simpler and there where greater priorities.

pfraze:
Hmm. Ok, let's merge this, but I think we need to keep working on the blob system as our top protocol priority (after the campjs local-discovery push). It's still pretty vulnerable.

Just as an example, if my feed published 1000 messages linking to unique 1gb files, what would alert us to the problem? (probably, it would be our pubs dying.) How long would it take us to deal with it? How many users would be affected? How easy would it be to help them?

The minutiae of what needs to be done includes:

 1. Size-limits on individual blobs (PW enforces 5mb in the composer, SSB does not)
 2. Configurable quotas on followed users (even friends can make mistakes or be compromised)
 3. Garbage collection
 4. Intervention + admin tools (eg delete blob x, list current blob sizes and quota usage, etc)
 5. Prioritization (eg "only get most-recent profile pics and image-posts on the feed in the last 5 days")

But I say this is a good thing! It's a big-idea project (not an "oh no let's come up w/something" project). We've got a pretty clear picture of our requirements at this point (at least, to make Patchwork work). And, this is the second really good application of our trust-network research, after spam-prevention.

The opportunity is to make something that's more reliable than Bittorrent (loses files all the time due to lack of persistent seeders) or the Web (which is host-dependent). What I think it'll broadly require is, a form of SLAs between peers to make the system reliable and predictable, and application-level metaknowledge to prioritize and garbage-collect.

dominictarr:
Okay most recent changes adds configuration support. I've just stored it in an array.

dominictarr:
@pfraze hmm, I think this may not be integrating nicely with how patchwork detects a new image has arrived. can you review, and point to where in patchwork that happens?

pfraze:
@dominictarr what are you seeing? serving code is https://github.com/ssbc/patchwork/blob/master/app/http-server.js#L62-L67

dominictarr:
well I was seeing some images not load that i thought should be (not as broken images but just not there)

does patchwork have a thing to fix the broken links when it does load?

pfraze:
yeah it's kind of a long-poll, the request stays open until `blobs.want()` calls back, so youll see nothing until found and then the image will populate

dominictarr:
Okay that is what I have implemented...

Hmm, okay so blob replication is way too slow. I'm using this version of scuttlebot with my patchwork install and things are taking too long to show up. It's connecting to many pubs, but doesn't seem to be downloading very fast... I think this needs a bit more work to figure out what is going on.

pfraze:
ok, you want me to postpone review on this? if no, I'll review tonight (if you respond quickly) or tomorrow

dominictarr:
I think I'll need to just use this a bit to try to figure out what is going on - if you want to help,
link this into your patchwork and see if it changes the patchwork behavior.

pfraze:
ok

dominictarr:
okay this is now ready to merge.

pfraze:
:+1: looks good

PR #263, opened 2015-11-23
don't try to connect to remote server without wifi.

dominictarr:
This detects if you have a network interface, and doesn't try to connect to peers that you won't be able to reach.

I'm starting to think it's time to work on gossip strategies again. I looked at this a while back, but decided that it wasn't worthwhile because there was only my pub, but now several people are running pubs.

This is on top of the blobs refactor which I think we'll merge this week.


pfraze:
:+1:

PR #266, opened 2015-12-11
fix validation: .previous is optional in messages (add())

pfraze:
`.add()` was incorrectly always expecting `.previous`. This makes it optional.

### DIFF
--- a/lib/validators.js
+++ b/lib/validators.js
@@ -54,7 +54,7 @@ module.exports = valid({
       return MissingAttr(n, 'sequence', 'number')
 
     // .previous
-    if (!ref.isMsg(v.previous))
+    if (v.previous && !ref.isMsg(v.previous))
       return MissingAttr(n, 'previous', 'msgId')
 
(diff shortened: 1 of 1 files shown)


pfraze:
@dominictarr rfr

dominictarr:
previous isn't optional, it's just that it's null on the very first message. previous is requried if sequence > 1

pfraze:
Validation will fail in a different area, then (the feed's add validation). This is the RPC parameter validation. If there's any form of input that's valid (as in the first message) it should be allowed here.

dominictarr:
@pfraze true. it's probably still more honest to check `if(msg.sequence > 1 && isMsgId(msg.previous))`

pfraze:
Hmm, yes that seems worth doing

dominictarr:
okay, I saw this had been forgotton and merged it, with a better check that only applies when sequence is > 1

pfraze:
:+1:

PR #267, opened 2015-12-20
lighter more flexible client

dominictarr:
This change means the `sbot` client doesn't load the entire scuttlebot dep tree, making loads faster.
Also, it gets the client manifest from the file (which previous version actually didn't do)

That is helpful for development because you can test a new plugin in scuttlebot/patchwork and hit it directly via the `sbot` command.


pfraze:
:+1:

dominictarr:
merged into 7.4.2

Issue #274, opened 2015-12-22
Error: Cannot find module 'pull-paramap' 


cryptix:
```
cryptix@mon:~$  sbot server
module.js:339
    throw err;

Error: Cannot find module 'pull-paramap'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/home/cryptix/node_modules/scuttlebot/plugins/replicate.js:2:12)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Module.require (module.js:366:17)

```

I got this after updating to `scuttlebot@7.4.3`. I wiped `node_modules` and reinstalled but got the same error.

Workaround: `npm install pull-paramap` [$ npm ls](https://ipfs.io/ipfs/QmUTAmU86f935xh8b8mA5NnPLL1WkrcxU7zcBS4nQGTnUw)

pfraze:
yep, it's incorrectly in devDependencies. PRing a fix

PR #275, opened 2015-12-22
package.json - move pull-paramap from devdeps to normal deps

pfraze:
Fixes https://github.com/ssbc/scuttlebot/issues/274


ralphtheninja:
:+1:

PR #278, opened 2015-12-29
Fix blobs plugin documentation mixup (add & rm).

lachenmayer:
The documentation for the `blobs` plugin commands `add` and `rm` is a bit mixed up.


pfraze:
Hah, thanks!

PR #280, opened 2016-01-13
Handle IPv6 addresses

clehner:
This patch updates scuttlebot to correctly parse addresses with IPv6 hostnames, as in
https://github.com/ssbc/ssb-ref/commit/3e68d06149e19847e8bcb6ea8975c9227ca555c5 and
https://github.com/ssbc/ssb-ref/commit/8d5f7c3badc8e5fc48090e384d8fa5458619a0f7

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -73,10 +73,11 @@ exports.merge = function (a, b) {
 exports.parseAddress = function (e) {
   if(isString(e)) {
     var parts = e.split(':')
+    var id = parts.pop(), port = parts.pop(), host = parts.join(':')
     var e = {
-      host: parts[0],
-      port: +(parts[1] || DEFAULT_PORT),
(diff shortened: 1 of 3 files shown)


pfraze:
Ok good deal. Could you add a test-case for this in test/invite.js? I think you could overwrite the server's getAddress to return an IPv6.

clehner:
Added a test which should fail before 5232100 and pass after it

clehner:
CI test failed at `replicate social network for animals`. Re-run it?

pfraze:
There might be a race in that test. Restarted...

pfraze:
Tests passing. @dominictarr good to merge?

dominictarr:
merged into 7.4.6

dominictarr:
sorrf 7.4.7

PR #285, opened 2016-01-19
Nicer logging

pfraze:
This PR makes a bunch of updates to quite down the logs, significantly.

1. The replication success-report now just tells you how many feeds were updated, and how many new messages there were. If there were no new messages, then nothing is output.
2. All log entries now include the ID of the remote. It previously used `._sessid`, which I think is old code because I dont see any references for it anymore.
3. Connection errors no longer include a stack trace.
4. Added a new log level, 'notice', which is between 'info' and 'warn'. So the full list, in order of most verbose to least, is: info, notice, warn, error.
5. Moved the replication summary to 'notice', and left all other info logging where it was.
5. Added `logging.level` config, which lets you set the verboseness. It defaults to notice, which means info will not be included.

The end-effect is, by default, you only see updates in the log now. If you want to see things like connection events, you can do so with `--logging.level info`.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -61,7 +61,14 @@ require('ssb-client')(keys, {manifest: manifest,
   port: config.port, host: config.host||'localhost',
   key: keys.id
 }, function (err, rpc) {
-  if(err) throw err
+  if(err) {
+    if (/could not connect/.test(err.message)) {
+      console.log('Error: Could not connect to the scuttlebot server.')
+      console.log('Use the "server" command to start it.')
+      process.exit(1)
+    }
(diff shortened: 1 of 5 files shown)


pfraze:
Also, added a nicer error message to bin.js when the server connection fails:

```
Error: Could not connect to the scuttlebot server.
Use the "server" command to start it.
```

dominictarr:
great! merging!

dominictarr:
merged into 7.4.4

cryptix:
:heart:

Issue #284, opened 2016-01-17
Blobs.add doesnt output hash-id when used from the CLI


pfraze:
Test with:

```
echo "hello, world" | sbot blobs.add
```

Does not return any hash

dominictarr:
this would require the change (discussed previously) to add callbacks to one way streams on muxrpc.

pfraze:
Hmm yeah, how do you figure we should deal with this in the short term? Should we special case the muxrpcli for this method?

dominictarr:
this only happens when you do `blobs.add` without posting a message, see code for publish which has a special case to post attachments.

pfraze:
That's true, but I think it'd be nice to have blobs.add work this way. Would you object to me adding a special case for that?

dominictarr:
@pfraze go ahead!

PR #286, opened 2016-01-20
hack: override blobs.add CLI call, in bin.js, to get the hash output

pfraze:
@dominictarr we discussed this previously. This will make `sbot blobs.add` output the hash of the added file.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -88,8 +88,30 @@ require('ssb-client')(keys, {manifest: manifest,
     cb()
   }
 
+  // HACK
+  // we need to output the hash of blobs that are added via blobs.add
+  // because muxrpc doesnt support the `sink` callback yet, we need this manual override:
+  if (process.argv[2] === 'blobs.add')
+    return blobsAddOverride(rpc)
+
   // run commandline flow
   muxrpcli(process.argv.slice(2), manifest, rpc)
 })
 
+// HACK helper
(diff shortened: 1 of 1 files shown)


pfraze:
Solves https://github.com/ssbc/scuttlebot/issues/284

pfraze:
Merged master into this branch, which included the tests fix so tests are now passing

dominictarr:
oh i just realized that the bit that used to do this in publish got obliterated with muxrpcli

dominictarr:
https://github.com/ssbc/scuttlebot/blob/7260b52d03cf865e06e6673c7f865fdf93e2e4d4/bin.js#L181-L221

dominictarr:
is that more useful? then you can post a message with attachments via one command.
or is this just so that documentation is consistent? maybe what we ought to do is provide a thing that wraps the client and does this so that ssb-client can be consistent too.

pfraze:
> is that more useful?

Both seem useful to me. I like having blobs.add output the hash for me. So if we also want to do this I'd be down.

> maybe what we ought to do is provide a thing that wraps the client and does this so that ssb-client can be consistent too.

that might be a good idea, instead of in sbot here

dominictarr:
what about... if a value is a file (source stream) `blobs.add` that instead, and replace that stream with a link. for the cli, we'd check if that arg is a file (or `.` for stdin)

dominictarr:
merged this into 7.5.0 for now anyway.

PR #289, opened 2016-01-21
fix CPU spin out on over quota

dominictarr:
It was a really stupid error, had two things named `delay` so it was causing `setTimeout(fn, NaN)` (probably) anyway, cpu levels are now acceptable.


dominictarr:
merging...

pfraze:
:+1: cool will watch for other symptoms

PR #290, opened 2016-01-21
replicate peers on the local network, even if they are strangers

dominictarr:
This _should_ replicate strangers you find on the local network, but only the their first 100 messages (by currently hardcoded quotas - easy to change) also it should replicate their avatars.

@mixmix you are more likely to have two computers running pw next to each other, can you test this?

### DIFF
--- a/plugins/gossip.js
+++ b/plugins/gossip.js
@@ -42,7 +42,7 @@ Peers : [{
   announcers: [ID+], // append only set.
                      // (unique members, only adds to end),
                      // could change for {id: true}
-  type: 'pub'|'manual'|'local'
+  source: 'pub'|'manual'|'local'
 }]
 */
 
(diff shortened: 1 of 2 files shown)


pfraze:
@dominictarr I tested this on my local machine and got it working, after some fixes. Check the commits to see what I changed. Somethings to mention:

 - For some reason, `cat()` did not appear to successfully add the local peers unless I moved the `pull.values` before the `createFriendStream`. Something is weird about that.
 - I found duplicates from `createFriendStream`, not just the local peers, so I modified the comment to reflect that

This PR is important for enspiral retreat so I'm going to get the tests passing and merge it

pfraze:
published `7.5.1`

mixmix:
Sorry i didn't see this earlier. Sounds good. Exciting!

dominictarr:
@pfraze oh I know, it's because createFriendStream is live and so it doesn't end (to handle case where you learn about a new peer when you replicate a follow message)

pfraze:
@dominictarr ah good

PR #291, opened 2016-01-22
add time measurement to replication summary notice-log

pfraze:
Adds some time information so we can get an idea of the throughput:

```
note @pzI REPL @uRECWB4KIeKoNMis2UYWyB2aQPvWmS3OePQvBj2zClg=.ed25519 success
  "Feeds updated: 20, New messages: 28, in 0m31s"
```

### DIFF
--- a/plugins/replicate.js
+++ b/plugins/replicate.js
@@ -168,6 +168,7 @@ module.exports = {
       //this is the cli client, just ignore.
       if(rpc.id === sbot.id) return
 
+      var startTS = Date.now()
       sbot.emit('log:info', ['replicate', rpc.id, 'start'])
       sbot.emit('replicate:start', rpc)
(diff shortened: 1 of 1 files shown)


dominictarr:
hmm, this might not represent total throughput because the connection will stay open for 30 seconds of inactivity, so the time of the connection doesn't necessarily show the total rate... it would be better to have progress events output a messages/second and then display that somewhere.

pfraze:
Yeah we could definitely do something better. Do you think it's worth the time? This is still somewhat useful

dominictarr:
yes I think performance monitoring is definitely a good thing, but it will take some iterations to get good.

pfraze:
@dominictarr ok, merge or close, up to you

PR #298, opened 2016-02-02
Demo PR - Sandboxed plugins

pfraze:
Yesterday, on the team call, we talked about CouchDB's architecture, which runs user-scripts to generate views and define queries. Using Couch's file-attachments system, they have the entire Couch Apps platform.

That got me thinking about how we might do the same in Scuttlebot. Perhaps it would be an avenue for the Ultra apps?

This PR demonstrates a sandboxed plugin which could support that. (I extracted it from EAppEnv.) It adds a `plugins.spawn(path, manifest)` function, which takes the JS file to execute, and the JS file's muxrpc manifest. The JS runs in a sandboxed VM in a child process, and uses STDIO for muxrpc traffic with sbot.

This plugin sandbox should support all of the existing plugins; that is, I think we could port them to run this way. See the changes to `index.js` for a usage example within sbot. See `plugins/hello-world.js` for a plugin definition. If you checkout this `sandboxed-plugins` branch, run `./bin.js server` to see the demo output.

(This PR has demo code - it's not merge-ready.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ var fs         = require('fs')
 var cmdAliases = require('./lib/cli-cmd-aliases')
 var valid      = require('./lib/validators')
 var apidocs    = require('./lib/apidocs.js')
+var plugins    = require('./lib/plugins')
 
 function toBuffer(base64) {
   return new Buffer(base64.substring(0, base64.indexOf('.')), 'base64')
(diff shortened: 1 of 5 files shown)


dominictarr:
I would rather see this published as it's own module, with tests and a clear idea about what properties the sandboxes have, and then a small change to scuttlebot to use that module.
There is just too much complexity here, that is orthagonal to the sort of stuff that is already in scuttlebot.
Or, maybe it should be to secret-stack, since that is where the plugin architecture is defined.

pfraze:
I think I can do everything in this PR as a module that either secret-stack or sbot could import.

dominictarr:
:+1:

pfraze:
@dominictarr @clehner https://github.com/pfraze/process-sandbox

PR #310, opened 2016-03-03
Fix want() callback value

### LINKED ISSUE #309: Blobs.want() calls back with false if there's any waiting period
If you call `blobs.want()` on a blob, and it's not available, the callback will be queued, and then will be called when the blob is downloaded. (This is correct.) However, the callback is not given a `true` for the `has` parameter.

Example usage from patchwork's blob server:

```js
sbot.blobs.want(hash, function(err, has) {
  if (!has) return respond(res, 404, 'File not found')
  //...
  respondSource(res, sbot.blobs.get(hash), false)
})
```

If the blob *was* already present, then `has` is set to true. I think it also should be set to `true` if there's a waiting period, for consistency, and in case we ever put a timeout on `want` (in which case `false` would be a meaningful result for `has`).

(This API issue is why Patchwork will show broken image links sometimes. The `has` comes back false when the download completes, resulting in the 404 response.)

pfraze:
Fixes #309

### DIFF
--- a/plugins/blobs/replication.js
+++ b/plugins/blobs/replication.js
@@ -151,7 +151,7 @@ module.exports = function (sbot, opts, notify, quota) {
       sbot.blobs.add(job.id, function (err) {
         if(!err) {
           delete jobs[job.id]
-          job.cbs.forEach(function (cb) { if(cb) cb() })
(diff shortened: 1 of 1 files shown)


pfraze:
@dominictarr ready for review

dominictarr:
merged into 7.5.3

Issue #312, opened 2016-03-04
`sbot invite.addMe` does not work


myf:
according to the [docs](https://ssbc.github.io/scuttlebot/plugins/invite.html) `sbot invite.addMe` is an alias to `sbot invite.accept`. when i execute this command with a valid invitation code I got the following error:

> node sbot.js invite.addMe [omitting actual invite code]

```
{ message: 'this.accept is not a function',
  name: 'TypeError',
  stack: 'TypeError: this.accept is not a function
    at EventEmitter.<anonymous> (/home/myf/modules/scuttlebot/plugins/invite.js:120:21)
    at apply (/home/myf/modules/scuttlebot/node_modules/muxrpc-validation/index.js:173:15)
    at EventEmitter.<anonymous> (/home/myf/modules/scuttlebot/node_modules/muxrpc-validation/index.js:82:14)
    at EventEmitter.hooked (/home/myf/modules/scuttlebot/node_modules/hoox/index.js:10:15)
    at Object.PacketStream.request (/home/myf/modules/scuttlebot/node_modules/secret-stack/node_modules/muxrpc/index.js:102:27)
    at PacketStream._onrequest (/home/myf/modules/scuttlebot/node_modules/packet-stream/index.js:149:17)
    at PacketStream.write (/home/myf/modules/scuttlebot/node_modules/packet-stream/index.js:122:41)
    at /home/myf/modules/scuttlebot/node_modules/secret-stack/node_modules/muxrpc/pull-weird.js:55:15
    at /home/myf/modules/scuttlebot/node_modules/secret-stack/node_modules/pull-stream/sinks.js:15:33
    at /home/myf/modules/scuttlebot/node_modules/pull-stream/throughs.js:73:11' }
Alias for accept.
```

dominictarr:
does it work if you use `accept`?

myf:
@dominictarr yes

dominictarr:
okay lets just remove `addMe` then, @pfraze what do yo think?

myf:
ah i think the problem is the scope, one sec

myf:
it's this line here https://github.com/ssbc/scuttlebot/blob/master/plugins/invite.js#L120
where this has a different scope? this.invite.accept might make sense but i don't quite understand what's going on here

pfraze:
@myf's PR fixes it, but it might be a good idea to cut the function.

dominictarr:
[...]

PR #313, opened 2016-03-04
fix for #312 addMe does not work as an alias to invite

myf:


pfraze:
@dominictarr did we remove addMe?

dominictarr:
this is was fixed by removing addMe, just having `invite.accept` instead.

PR #314, opened 2016-03-05
disable all quotas by setting party mode

dominictarr:
quotas wasn't working well, and, I suspect that calculating the path to the sharer in the blobs quotas system might be what is causing the spinning cpu.

``` 
ssb_party=1 ./bin.js server
```
or edit config. otherwise this is backwards compatible, so we can merge this into patchwork and let people who where having problems test it an see if it improves.

### DIFF
--- a/plugins/blobs/replication.js
+++ b/plugins/blobs/replication.js
@@ -73,6 +73,7 @@ module.exports = function (sbot, opts, notify, quota) {
   }
 
   function limitFor(id) {
+    if(opts.party) return -1
     var h = hops(id)
     if(hops === -1) return conf.minLimit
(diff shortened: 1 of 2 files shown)


pfraze:
Ok :+1: this is a temporary feature. After a testing period, we'll either need to remove quotas or give this flag a getter name (party is unclear).

pfraze:
@dominictarr are you going to publish this?

pfraze:
oh yeah, `7.6.0`

dominictarr:
sorry, internets a bit sloppy right now, about to head into the library so
I can deal with all this http.

PR #315, opened 2016-03-06
set remote key via host, to be client to another sbot

dominictarr:
Once upon a time this was possible but it got trampled somewhere.
This enables you to set a `master` configuration on your pub, and command your pubserver to accept commands from your local client, which has better latency than sshing onto the server and then running local commands.


pfraze:
Looks :+1: to me

PR #321, opened 2016-03-18
Add comments, do minor cleanup

pfraze:
Adds commenting and removes some dead code

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -2,101 +2,106 @@
 
 var fs           = require('fs')
 var path         = require('path')
-var msgs         = require('ssb-msgs')
 var pull         = require('pull-stream')
 var toPull       = require('stream-to-pull-stream')
 var explain      = require('explain-error')
 var ssbKeys      = require('ssb-keys')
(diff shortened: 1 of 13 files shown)


bin.js
```diff
@@ -2,101 +2,106 @@
...
     }
-    throw err
-  }
 
-  // add aliases
-  for (var k in cmdAliases) {
-    rpc[k] = rpc[cmdAliases[k]]
-    manifest[k] = manifest[cmdAliases[k]]
-  }
+    // add aliases
```
dominictarr: should not add extra indentation here, indentation was already aligned.

dominictarr:
merged into 7.6.6

bin.js
pfraze: I did it to balance the code-paths, from

```js
if (server command) {
  code path 1
  return
}
code path 2
```

to

```js
if (server command) {
  code path 1
} else {
  code path 2
}
```

which I think is easier to scan, but idk if you change it back

PR #323, opened 2016-03-24
if an invite code is surrounded in quotes, automatically remove them

pfraze:
Invite codes are outputted with quotes around them, so sometimes they get shared that way. I had somebody get tripped up by this, so automatically stripping the quotes is probably a good idea.


dominictarr:
merged into 7.6.7, with some added tests.

PR #327, opened 2016-03-28
Tracking PR: Userland Plugins

pfraze:
Adds two CLI-only commands: `install` and `uninstall`.

The install shells out to npm, and installs the module to `~/.ssb/node_modules` using `--global-style`, for good dep isolation and atomic installs.

On server-load, sbot will scan the modules, check their signature, and attempt to `.use()` them.

Tracking PR: Needs testing and automated tests, and we need to make sure we're :+1: on the implementation.


dominictarr:
can user-plugins be a plugin? if it took config as the second argument it would have the right signature.

dominictarr:
maybe we'll need to change secret-stack slightly to allow plugins to be loaded after initialization.

dominictarr:
and maybe we need a `sbot restart` command too... which will require a few changes to `sbot server`

pfraze:
Why plugin vs this?

I did that originally but it's harder to do good log output that way because of the constraints of our muxrpcli setup. It also meant `sbot plugins.install` instead of `sbot install` which isnt as nice (but Im ok with it)

(If you look in the commit history, you can see how I did it.)

pfraze:
I checked into using npm programmatically instead of shelling it out. It worked, but it (apparently?) involved including npm as a dep. require('npm') didnt work unless I did.

That seemed hard to stomach at the moment, but now that I think about it, it might be necessary in the electron version.

dominictarr:
lets not include npm as a dep. that will just be annoying...
I think if you make install a source stream, and pipe the output of npm staight out, then it will just work.

I think we should also have `enable` and `disable` configs, so you can turn plugins on and off (for example, disable buggy plugin when not testing it. Then use this instead of just loading everything in the directory.

or maybe `--plugin.name` to enable, `--no-plugin.name` to disable.

pfraze:
@dominictarr maybe `enable` and `disable` update the `~/.ssb/config` so that it has the same effect as `--plugin.name` and `--no-plugin.name`

pfraze:
changed to a plugin

pfraze:
Added enable and disable

@dominictarr rfr again

pfraze:
Added some tests.

Note, I updated the definition of `.close()` in scuttlebot to close the SSB database. It previously wasnt closing the SSB database, which causes problems when a test needs to reuse a data directory across multiple tests (which plugins does).

plugins/plugins.js
```diff
@@ -0,0 +1,172 @@
...
+            // - use basename(), because plugins can be installed from the FS, in which case pluginName is a path
+            var name = path.basename(pluginName)
+            config.plugins[name] = true
+            writePluginConfig(name, true)
+            p.end()
+          })
+        return cat([
+          pull.values([new Buffer('Installing "'+pluginName+'"...\n', 'utf-8')]),
+          toPull(child.stdout),
+          toPull(child.stderr),
```
dominictarr: instead of concatenating the stdout and stderr it would be better to use pull-many so they come out the same as you'd normally see it.

`cat([installing, many(stdout, stderr), p])`

dominictarr: and if the install fails, we should make the stream error,
which you can do by calling `pushable.end(error)`
https://github.com/dominictarr/pull-pushable/blob/master/index.js#L40-L43

dominictarr:
This is looking pretty good.

hmm, I'm thinking we should make sbot command hit a bash script first, and handle the sbot server command inside a bash loop, then we can add a sbot restart command. patchwork would need automatic reconnections, but that is a good idea anyway ;)

dominictarr:
hmm, what if we refactored all the ssb plugins, and but kept the "core" ones as dependencies of scuttlebutt. we'd try resolving packages from `.ssb/node_module/` first, and if it isn't there, then we go to `scuttlebot/node_modules/`, if it's not there either, we failed to load it.

pfraze:
> hmm, I'm thinking we should make sbot command hit a bash script first, and handle the sbot server command inside a bash loop, then we can add a sbot restart command. 

So then we could send a kill signal to the sbot process to do reboot, and to the bash process to close, correct?

> patchwork would need automatic reconnections, but that is a good idea anyway ;)

I want to rewrite patchwork as a plugin, so its backend will automatically reboot, and then, yeah, the frontend can just poll for the WebSocket to be restored.

> hmm, what if we refactored all the ssb plugins, and but kept the "core" ones as dependencies of scuttlebutt

Sounds good to me. [related?](https://twitter.com/dominictarr/status/715088574576656384)

pfraze:
I made your suggested code updates.

Suggest we merge as-is, then begin factoring out plugins and making the restartable server in separate PRs. :+1: ?

clehner:
How do I test this? I've checked out the plugins branch but can't run plugin commands. 
`./bin.js plugins.install` or `./bin.js install` show Command not found, and the `./bin.js` help message doesn't show the plugins plugin.

pfraze:
@clehner hmm, are you sure you're running the sbot server from this branch?

clehner:
@pfraze that was it. I configured patchwork to use the plugins plugin and am able to use the plugins plugin.
A couple things I found:
- When I first installed a plugin, the install command completed successfully, but there was no ~/.ssb/node_modules directory. After I created that directory and re-ran install, the plugin got installed into it.
- I ran `./bin.js plugins.install pfraze/ssb-notifier#pluginify` to install ssb-notifier from your pluginify branch, and got the success message. However, `~/.ssb/config` was updated with the plugin name including the #rev part, which I'm guessing shouldn't be there:
```
"plugins": {
    "ssb-notifier#pluginify": true
  }
```

- After editing the config to have `"ssb-config": true`, it seems that the plugin is still not getting loaded. I edited `~/.ssb/node_modules/ssb-notifier/server.js` to produce a notification when it is loaded, but I don't see the notification.

pfraze:
First two bullets are unsurprising, thanks for the heads up. About bullet three, did you change it to `"ssb-config": true`, or `"ssb-notifier": true`? If you did the former, that's your mistake -- it should be the latter.

clehner:
> About bullet three, did you change it to "ssb-config": true, or "ssb-notifier": true

My bad - it's the latter. The config is in fact:
```
{
  "party": true,
  "allowPrivate": true,
  "plugins": {
    "ssb-notifier": true
  }
}
```

pfraze:
Did you restart the server after updating config? Does the server log tell you if it's loading the plugin? It should tell you its decision for each item in your node_modules.

pfraze:
(And to revise my statement, actually, it not working till node_modules is created *does* surprise me.)

clehner:
Ok, running sbot server I saw the message about the plugin loaded, and it was indeed loaded. I changed patchwork/index.js to make it call loadUserPlugins and now the plugin loads in patchwork too. So it all seems to be working now. Carry on!

pfraze:
ok cool, I'll make some revisions based on what you caught and push

pfraze:
Hmm. Actually a bit tricky to know what the module name will end up being, given [how many different schemes that npm install supports](https://docs.npmjs.com/cli/install). I think I'll have to watch the output of the install command, and pull it out? That sounds ugly.

pfraze:
Bleh. I think I'll have to use `--json` so I can get the package out. There goes the natural npm output. Weak.

pfraze:
That appears to work as expected.

dominictarr:
oh yeah, some of these is the reason that npm is not really ideal for this purpose. like, I can't install a module with a different name, so forks can be weird. But, since it's agreed this is a temporary experiment, that is okay.

pfraze:
Yeah, if you fork you'll have to change the package name.

Ok, are we ready to merge this?

dominictarr:
testing this branch I tried to install ssb-links and it installed, and seems to load, but the cli gives me an error that there is no command "links2.dump" and it does not appear in the manifest.

dominictarr:
oh, no it was a problem with links2! okay I'll do some more testing but looking okay

dominictarr:
I installed ssb-links, then ssb-query, and got this output:
```
plugins.install ssb-query
Installing "ssb-query"...
"ssb-links@2.0.0" has been installed. Restart Scuttlebot server to enable the plugin.
```
but it does seem that ssb-query has been installed. you should probably dump more npm output. npm is surprising.

pfraze:
Ah, I was afraid that might happen. NPM's --json output just dumps out an object that looks like this:

```js
{
  dependencies: {
    packageName: {...},
    packageName: {...},
    packageName: {...)
  }
}
```

I was hoping there'd be only one `packageName` in that dependencies object (the one you installed) but I'm betting that ssb-links was in there too.

Honestly, NPM, stop *trying* to make me hate you.

I'll poke at this more today.

pfraze:
Yeah, it's worse than I realized.
 1. That --json output actually lists the contents of your node_modules directory. So, we'd have to capture a before+after and do a diff.
 2. The leveldown install emits into stdout, so now the output needs sanitizing before parsing into JSON. That won't be easy, since it's pretty unpredictable what the modules' install scripts will dump out.

Time to pull out a classic:

fuuuuuu

pfraze:
Here's what I suggest.

Let's not bother capturing the NPM output. It's exceeding the effort/reward curve threshold. 

Instead, let's change the `install()` parameters so that it looks like this:

```bash
sbot plugins.install {pluginName} [--from path]
```

So the `pluginName` must always be *just* the name of the plugin, and `{from: path}` is an optional override of where to install from.

The default behavior:

```bash
sbot plugins.install ssb-links
```

Would install latest from npm. And we can probably manage to support versions, too:

```bash
sbot plugins.install ssb-links@1.0.0
```

But, if you want to install from somewhere else, you do:

```bash
sbot plugins.install ssb-links --from ~/my-ssb-links
sbot plugins.install ssb-links --from https://github.com/dominictar/ssb-links
etc
```

pfraze:
Got `--from` working in another branch, which I'll merge here if yall 👍 

One thing was, it doesn't let you choose a *different* name from the plugin, as you might expect. For instance:

```bash
sbot plugins.install ssb-not-notifier ~/ssb-notifier
```

That will cause the `~/.ssb/config` to list `ssb-not-notifier`, but it'll install to `~/.ssb/node_modules/ssb-notifier`.

It might be fixable by installing to a tmp directory, then moving the folder into `~/.ssb/node_modules` with the altered name, but I suggest we just dont do that either, for now.

dominictarr:
@pfraze :+1: that also gives us a way to do an `plugins.install module as other_name` thing

pfraze:
@dominictarr Ok so you figure I should go ahead with

> installing to a tmp directory, then moving the folder into ~/.ssb/node_modules with the altered name

dominictarr:
even simpler, and then we also get the full npm install output back, right?

pfraze:
Yeah, correct

pfraze:
Ok we're :metal:

clehner:
When I try to install `ssb-links` or `ssb-query` as a `scuttlebot@plugins` plugin, I get a crash:

```
$ sbot plugins.install ssb-query
Installing "ssb-query"...

> leveldown@1.4.4 install /tmp/node_modules/ssb-query/node_modules/streamview-links/node_modules/level/node_modules/leveldown
> prebuild --install

ssb-query@0.0.2 ../node_modules/ssb-query
├── explain-error@1.0.1
└── streamview-links@2.0.2 (ltgt@2.1.2, pull-paramap@1.1.4, pull-write@1.0.2, pull-sink-through@0.0.0, pull-stream@3.3.0, map-filter-reduce@2.2.1, bytewise@1.1.0, pull-level@1.4.2, level@1.4.0)
Error: unexpected end of parent stream
```
scuttlebot:
```
fs.js:808
  return binding.readdir(pathModule._makeLong(path));
                 ^

Error: ENOENT: no such file or directory, scandir '/tmp/ssb-query/node_modules'
    at Error (native)
    at Object.fs.readdirSync (fs.js:808:18)
    at ChildProcess.<anonymous> (/usr/local/src/scuttlebot/plugins/plugins.js:122:29)
    at emitTwo (events.js:87:13)
    at ChildProcess.emit (events.js:172:7)
    at maybeClose (internal/child_process.js:821:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:211:5)
```
When I look at `/tmp/ssb-query` after this, I find it is an empty directory.

@pfraze could the build step of leveldown be getting in the way of scuttlebot's parsing of npm's output?

pfraze:
No, it doesnt parse the stdout anymore. Something must be broken.

clehner:
I tried installing those plugins again and the installation worked without error. I'll try to investigate more if that ENOENT happens again. Is there anything else blocking `plugins` from being merged?

pfraze:
Let's discuss it on the talk tomorrow

clehner:
@pfraze @dominictarr can one of you merge and publish this? I would like to use it downstream

pfraze:
Soon as Im off the clock on my client work, I will

pfraze:
published `8.1.0`

pfraze:
(I said that prematurely, the tests had to run - again - before publish went through. NOW published.)

arj03:
This is pretty exciting!

PR #330, opened 2016-03-28
Wrong check for previous in first message

arj03:


pfraze:
Hate to say it but this is wrong, the sequence starts at 1, not 0.

dominictarr:
the sequence starts at 1 so that, during replication you can send 0 to signify that you don't have anything in that feed but want it. sending 1 means you already have the first message, etc.

arj03:
Okay, but I ran into this problem when I imported a whole feed from files using https://github.com/arj03/ssb-exporter. The first message will fail to import as it is without this PR.

pfraze:
@arj03 you updated the PR to check if sequence == 1. Does the import fail now?

arj03:
It still works, so this is the correct patch as far as I can see and test. Do you want me to do a new PR or is it okay with two patches?

pfraze:
Nah this is :+1:

pfraze:
Published `7.6.9`

PR #339, opened 2016-04-15
Point API reference to fixed URL

dangerousbeans:


pfraze:
Thanks, actually we should update to use the site we recently put together, https://scuttlebot.io/apis/scuttlebot/ssb.html

PR #343, opened 2016-04-24
Expose log command in logging plugin

clehner:
This makes the log function from the logging plugin be usable as a method/command. This makes it possible for ssb apps connecting to scuttlebot using ssb-client to log messages to the same output to which sbot's log messages go.

Also I added an API doc for the log function and updated an isString function to be more accepting of strings

### DIFF
--- a/lib/apidocs.js
+++ b/lib/apidocs.js
@@ -8,5 +8,6 @@ module.exports = {
   gossip: fs.readFileSync(path.join(__dirname, '../plugins/gossip.md'), 'utf-8'),
   invite: fs.readFileSync(path.join(__dirname, '../plugins/invite.md'), 'utf-8'),
   'private': fs.readFileSync(path.join(__dirname, '../plugins/private.md'), 'utf-8'),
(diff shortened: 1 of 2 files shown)


pfraze:
Is this a good idea? Shouldnt apps have their own logs?

clehner:
> Shouldnt apps have their own logs?

Maybe so. I wanted to reuse the functionality of logging with levels. An app running as a plugin would have access to the server's log. There isn't a feature-checking way to check if that log is available though because it is done as event listeners, which is why I added the log method.

However, if scuttlebot logging is meant to be internal to scuttlebot, an app can still get the logging level from the config, and the levels functionality is simple enough, so this isn't a big need.

clehner:
Wouldn't centralizing the logs allow for fixing the problem of the new gossip logging not being configurable? (mentioned on GH or PW, I forgot)

pfraze:
No, that's just happening because the latest logging code is using console.log() instead of emitting log messages

PR #348, opened 2016-06-09
refactor out blobs into ssb-blobs plugin

dominictarr:
This removes all the ssb-blobs stuff,
and blobs is now an independent plugin.

the best test is to use this to send and receive encrypted files:

you'll need to install `ssb-blobs` plugin.

clone this repo on `ssb-blobs` branch, and install ssb-blobs

```
sbot plugins.install ssb-blobs
# restart your scuttlebot server, running from `./bin.js server` in this repo.
```
install `ssb-secret-blob`:
```
npm install -g ssb-secret-blob
```

then, post a blob, as described in [sblob docs](https://github.com/ssbc/ssb-secret-blob)

`sblob box {filename}`
which will output a hash + a key
and then open that
`sblob unbox "{id#key}"`

to try one i prepared earlier:

```
sblob unbox '&Ux+KG30u/wwlo08oRjp++C0sAj0hGJYnzKgv1fuAnv8=.sha256#QL8h3iEt3HFNOcCCumdxqrUKntM5wCKef53Sl+tKqlM='
```


dominictarr:
@mixmix and I tested this and it worked great over local wifi, but not so well over the pubs (possible that is because it's currently only my pub which can do this, so it takes a while)

Issue #349, opened 2016-06-15
gossip on wakeup or net change


dominictarr:
Just realized that after the gossip refactor, I forgot to include gossip on wakeup.
https://github.com/ssbc/scuttlebot/blob/master/plugins/gossip/schedule.js#L3

is still there, but nothing uses it.

dominictarr:
looking more into this, I realized that the problem is with keep-alive. If I connect a TCP connection to a server, and then change networks, node doesn't realize that the connection is dead. I heartbeats are needed... I have been thinking about a new version of pull-box-stream that has integrated heart beats (a packet could have an expect next packet time...)

plan to fix (short term): add reconnect command that forcibly disconnects peers (thus triggering reconnect)

plan for long term: replace pull-box-stream.

PR #350, opened 2016-06-18
reconnect after change network or suspend

dominictarr:
This should fix https://github.com/ssbc/scuttlebot/issues/349
gonna test suspend while I got get a bite. then merge if this works.


dominictarr:
okay this seems to work

PR #353, opened 2016-06-26
Fix 11f01217 to make 0f96771a work again

kyrias:
Commit 11f01217 (oops, fix so you can actually start the server when not
using --, 2016-05-15) broke the change from commit 0f96771a (separate
argv by -- and use later part for configuration., 2016-05-14).

Fix this by adding back the i part of the argv.slice.

Signed-off-by: Johannes Löthberg <johannes@kyriasis.com>

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -17,7 +17,7 @@ var cmdAliases   = require('./lib/cli-cmd-aliases')
 var argv = process.argv.slice(2)
 var i = argv.indexOf('--')
 var conf = argv.slice(i+1)
-argv = ~i ? argv.slice(0) : argv
+argv = ~i ? argv.slice(0, i) : argv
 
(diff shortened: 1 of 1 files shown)


pfraze:
Looks good, @dominictarr good to merge?

dominictarr:
@kyrias thanks - I am on 8.3.2 (1c0b648aed5127e573b76df37940e9d874fe75c2)
and can start the server fine, how do you reproduce the problem?

kyrias:
The problem isn't with starting the server, it's that the second commit broke passing config items to all commands in the commandline using `--` as a separator. 
So things like `sbot invite.create 1 -- --path foo --port 666` no longer worked.

kyrias:
Ping?

dominictarr:
sorry, thanks for reminding me

kyrias:
Awesome, thanks.

PR #357, opened 2016-09-26
Add onion support to invite plugin

arj03:


dominictarr:
lets avoid inlining special cases like this. we also had a thing to convert a legacy format address into multiserver, lets move that into a module and then just use that module where we need it.

PR #367, opened 2016-12-13
pass caps from config to ssb-client

cryptix:
before, the ` "caps": {"shs": "...." } ` in `.ssb/config` wouldn't get passed to `sbot` client calls, like `whoami`.

(re %sRD2wfvldRwi95yqTTN5SRnmoG0c1ktcXFSUF8lQlh8=.sha256)


cryptix:
test failure due to #365

dominictarr:
merged into 9.4.1

PR #371, opened 2017-02-10
[flume] fix bin export in package.json

ahdinosaur:
also export `scuttlebot` command, since `sbot` can be confusing if you expect the convention where the exported command matches the name of the package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,8 @@
     "tape": "~4.0.0"
   },
   "bin": {
-    "sbot": "./sbot.js"
+    "sbot": "./bin.js",
+    "scuttlebot": "./bin.js"
   },
   "scripts": {
     "prepublish": "npm ls && npm test && noderify bin.js > sbot.js",


dominictarr:
I think this should be discussed a bit more here: %BIn4rQl8o7PxeolhiZCw77fUN5gGpOv6I3gDXhMfP+M=.sha256

mmckegg:
I think this should be merged! @ahdinosaur @dominictarr

This is such a simple fix that would have saved me at least 5 mins the first time I tried to run it! 

I **absolutely detest** packages that name the bin command differently than the package name. This PR solved the confusing aspect and didn't remove anything from people who like short commands.

But IMO shorter global commands aren't necessary better. More likely to conflict with something else, or are harder to figure out what they do, (or what package they come from). A user that likes short commands is also probably capable of create aliases for these!

/rant 🤣

dominictarr:
you've persuaded me @mmckegg merging.

ahdinosaur:
looks like this was manually merged: https://github.com/ssbc/scuttlebot/commit/4a614609e8f7ecc7dcbf5de0a4e156d065f495fd

PR #374, opened 2017-03-09
Improve replication error handling

clehner:
as discussed in [%SMqHQTSeyR8Zq0KcSe2z0HbvjW1CqKIlJ1leuVVL4G0=.sha256](https://viewer.scuttlebot.io/%25SMqHQTSeyR8Zq0KcSe2z0HbvjW1CqKIlJ1leuVVL4G0%3D.sha256):

- log a stream error only if it is not one of the expected benign errors, and then only log it once instead of logging it for each substream.

also, pre-emptively close legacy erroring substreams, to save bandwidth (analysis in [%368B7blWHcWTEzqGO4DDipDBCLwvRlKxwVxTOFtV9kY=.sha256](https://viewer.scuttlebot.io/%25368B7blWHcWTEzqGO4DDipDBCLwvRlKxwVxTOFtV9kY%3D.sha256)

and also, show Connected/Disconnect on the log for peers that are not in the peer table (e.g. incoming non-pub connections)

### DIFF
--- a/plugins/gossip/index.js
+++ b/plugins/gossip/index.js
@@ -176,10 +176,12 @@ module.exports = {
       //don't track clients that connect, but arn't considered peers.
       //maybe we should though?
       if(!peer) {
-        console.log('Connected', rpc.id)
-        rpc.on('closed', function () {
-          console.log('Disconnected', rpc.id)
-        })
+        if(rpc.id !== server.id) {
+          console.log('Connected', rpc.id)
+          rpc.on('closed', function () {
+            console.log('Disconnected', rpc.id)
+          })
+        }
         return
       }
 
--- a/plugins/gossip/index.js
+++ b/plugins/gossip/index.js
@@ -112,7 +112,7 @@ module.exports = {
         peer.state = 'disconnecting'
         peer.stateChange = Date.now()
         if(!peer || !peer.disconnect) cb && cb()
-        else peer.disconnect(null, function (err) {
+        else peer.disconnect(true, function (err) {
           peer.stateChange = Date.now()
         })
 
--- a/plugins/gossip/index.js
+++ b/plugins/gossip/index.js
@@ -175,7 +175,14 @@ module.exports = {
       var peer = getPeer(rpc.id)
       //don't track clients that connect, but arn't considered peers.
       //maybe we should though?
-      if(!peer) return
+      if(!peer) {
+        console.log('Connected', rpc.id)
+        rpc.on('closed', function () {
+          console.log('Disconnected', rpc.id)
+        })
+        return
+      }
+
       console.log('Connected', stringify(peer))
       //means that we have created this connection, not received it.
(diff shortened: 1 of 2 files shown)


dominictarr:
hey sorry, my reply didn't stick before. I was wondering if keeping the error message on the peer object would be useful? some of the errors (such as EHOSTUNREACH tell you that the server doesn't exist, not just that the connection was terminated, which is useful information) otherwise this looks good to merge

clehner:
the errors handled in this code are mostly due to the rpc stream terminating. errors coming from internal to createWriteStream could also be caught here, but i haven't seen any. i added to the streamErrors object the ones that i saw, since none seemed serious enough to warrant falling through to be reported to the user on the console. i don't see these errors as being very useful to save on the peer object, but they could be used perhaps for various things. that would be separate to this PR, but here are my thoughts about the errors i saw and what i could see them being useful for, if anything:

- 'read EHOSTUNREACH': i'm not sure exactly how this error happened, but i don't think it could have been that the server didn't exist, because then the tcp connection and secret handshake would not have succeeded. this error came from the 'read' syscall. i would guess it was that the network state changed such that packets could no longer be sent on the connection. that might or might not be particular to that peer though. so this could be used to infer that we can't reconnect to that peer now, but i'm not sure how we would detect when it would be safe to reconnect. maybe put some time delay on it in the gossip schedule, to avoid connecting to it "for a while", relative to other peers.

- ENETDOWN: could be used to tell that the network interface went down and so making outgoing connections could be paused. on-change-network might already pick up on this though; more testing could be done to see if this is the case.

- ETIMEDOUT could be used to infer that the peer is behind a slow link and should be deprioritized - but it could also be a local network issue. if info was kept track of for each peer how many connections with it closed due to timeout (with recent connections weighted more), that could useful to feed into the gossip scheduling, as factor in a weighted random pick.

- ECONNRESET, EPIPE, unexpected hangup: could be a broken connection, or could be the peer destroyed the connection, i think. not sure what could be usefully inferred from this.

- unexpected end of parent stream && !(err instanceof Error): indicates the peer has packet-stream v2.0.0. could be useful to know for future connections if/when we have multiple replication methods to choose from, since we could mark this peer as a legacy peer - but we probably already know this since we had to make a connection in order to get this error.

Also, the duration of the connection could be useful. if we know that a peer's connections often end after a few seconds, maybe we can use that to decide to delay requesting replication after the connection is opened, to avoid wasting bandwidth.

For gossip scheduleing, more useful errors than these may be the ones received when attempting to connect to peers.

dominictarr:
@clehner very good point. if the peer doesn't exist, EHOSTUNREACH at the callback to `connect(peer, cb(err, rpc){})` it might be a good idea to track that error, but this PR is not the place to make that change.
gonna merge this then.

dominictarr:
merging into 9.4.6

dominictarr:
merging into 9.4.6

PR #375, opened 2017-03-12
Adding docker support

DeveloppSoft:
The following pull adds a simple Dockerfile to use with docker.
There is some doc inside.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,15 +16,14 @@
 FROM node:alpine
 MAINTAINER DeveloppSoft
 
+EXPOSE 8008
+
 USER root
 
 RUN apk update
-RUN apk add python python-dev alpine-sdk libsodium-dev libsodium
+RUN apk add python python-dev alpine-sdk
 
 RUN npm install scuttlebot@latest -g
 RUN npm install secure-scuttlebutt -g
-
-# Install plugins
-RUN npm install ssb-links ssb-query ssb-ws ssb-fulltext
 
 CMD /bin/ash


dominictarr:
I don't think it needs `apk ... libsodium` because npm completely handles that.
also, what is alpine-sdk?

Oh, there are also some other docker scripts https://github.com/ktorn/docker-scuttlebot
and https://github.com/staltz/easy-ssb-pub also has a dockerfile. I guess we should move one of these into ssbc so that it's easy to find. @ktorn @staltz what do you think?

DeveloppSoft:
`alpine-sdk` is a nice package of basic development tools.
I had problems if `python` and `python-dev` weren't installed.
I agree `libsodium` could be removed.

yangwao:
Yes, python and python-dev is required.

dominictarr:
is this still relevant? can it be closed, still waiting to hear from @ktorn and @staltz

ktorn:
@dominictarr thanks for the nudge.

Until now I preferred the minimalist approach, which I'm still using on my own pub.

If I had to vote on one to merge to the main project, I'd vote for the @staltz code , which does add the easy web-based invite feature.

DeveloppSoft:
The one from @staltz is quite interesting (I live the web invite feature).
On the other hand, the current Dockerfile was designed to run a simple node/pub, it is based on Alpine Linux to reduce its size.
The best option would be to have easy-ssb-web in an Alpine container (last time I checked, it was using Debian).

dominictarr:
@staltz's thing adds a bunch of stuff. I don't really know docker but since it's just one file I guess I can merge it.

okay looking at the code, I don't think the plugin install will work. those need to be installed into `~/.ssb/node_modules` not just anywhere. Also, it looks like this doesn't actually start sbot? is that correct?

DeveloppSoft:
Indeed.
I can clean the file a bit (removing plugins and some deps).
You are true, it doesn't start sbot, to let you do what you want, what can find some docs in the header (copy pasting here)
```
# To build:
#    docker build -t sbot .
#
# The following allows you to start the server:
#    docker run -it --rm --port 8008:8008 sbot sbot server
#
# To start an isolated network, run many times the following (no ports, exposed):
#     docker run -it --rm sbot sbot server
# You may also need to use the first command (only once)
#
# To get a shell:
#     docker run -it --rm sbot
#
```

staltz:
i think this issue can be closed since a docker config can be published as a separate package from this one, like easy-ssb-pub does. 

I'd also vote for *not yet* migrating easy-ssb-pub to the SSBC org because, as dominic said, it does a bunch of stuff, and is opinionated. SSBC needs unopinionated modules which are super essential as building blocks for programmers. It's also easier in the SSBC if we don't need to argue with logic like "because I want it to be like this", so we can stick to technical argumentation. I'm trying to build easy-ssb-pub as a collection of SSB stuff packaged together for non-programmers (the goal is to do away with terminal commands entirely, somehow) to make it easy to host a pub without any special knowledge. Also, as I'm making up my mind with easy-ssb-pub, I'm going to refactor some parts and publish them as ssbc modules.

dominictarr:
@staltz interesting points. I think that makes sense from an adoption perspective, but personally I'm more concerned right now with a developer on boarding perspective, and if docker helps with that, then we can have a simple docker script on ssbc. I've used docker once but I'm personally fine with just npm.

staltz:
True, we need to help developers on board. So you mean simple docker script for sbot only?

dominictarr:
yeah, I understand that is the intention here, I think that's what @ktorn's is about too.

staltz:
Oh, my bad, I thought this was an issue but it's actually a PR when I said:

> i think this issue can be closed since a docker config can be published as a separate package from this one, like easy-ssb-pub does.

DeveloppSoft:
OK, just pushed some changes

christianbundy:
I'm looking to see whether an official Dockerfile for sbot exists, but after reading through the above discussion I'm not sure I understand where we're at. Can anyone comment on the state of this PR? Thanks!

ahdinosaur:
hi @christianbundy, i'm not sure where this pull request is at, but i can say i'm working on Docker-powered pubs with [PeachCloud](http://peachcloud.org) and that [`ssb-pub`](https://github.com/ahdinosaur/ssb-pub) and [`peachpub-peer`](https://github.com/peachcloud/peachpub-peer) have usable `Dockerfile`s, happy to help if i can.

PR #376, opened 2017-03-13
fix realtime replication push

mmckegg:
Finally tracked this thing down! 🎉 

There is a typo in `plugins/replicate.js` that uses `opts.sequence` instead of  `opts.sequence || opts.seq` causing realtime push to break.

This is most commonly noticeable when connected to someone on the local network and you don't see their posts until after they restart, but I suspect this has been breaking persistent gossip for ages!

cc @dominictarr @mixmix @ahdinosaur @clehner

### DIFF
--- a/plugins/replicate.js
+++ b/plugins/replicate.js
@@ -99,7 +99,7 @@ module.exports = {
           pushable.splice(i, 1)
         })
         pushable.push(p)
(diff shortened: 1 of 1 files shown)


dominictarr:
merged into 9.4.5! thanks

mixmix:
Arrrrg. Nice one!

PR #379, opened 2017-03-23
fix bug replicating with local peers

ahdinosaur:
taken from https://github.com/ssbc/patchwork/commit/0b9adb1dfcfd2248f55f9dc2130f2d53aeb91794

@mmckegg and i have no idea _exactly_ why an error happens without this, only that it _somehow_ _sometimes_ fixes _something_. yikes. :sweat_smile:

### DIFF
--- a/plugins/replicate.js
+++ b/plugins/replicate.js
@@ -127,11 +127,13 @@ module.exports = {
 
     function localPeers () {
       if(!sbot.gossip) return
-      sbot.gossip.peers()
-        .forEach(function (e) {
-          if(to_send[e.key] == null)
-            addPeer({id: e.key, sequence: 0})
-        })
+      sbot.gossip.peers().forEach(function (e) {
+        if (e.source === 'local' && to_send[e.key] == null) {
+          sbot.latestSequence(e.key, function (err, seq) {
+            addPeer({id: e.key, sequence: err ? 0 : toSeq(seq)})
+          })
+        }
+      })
     }
 
     //also request local peers.


dominictarr:
Hmm, this would have fired for pub servers before.
I would merge this, but it is more important to understand the problem.

Can you tell me about the train of thought that led here? how did you observe the problem? where did you see the error? did you try anything else before you tried this?

I am quite puzzeled how "wrong number" is an error you see here, it shouldn't even get to replication. that is when a peer tries to connect to you, but has the wrong public key.

One guess i have is if invalid arguments are sent to createHistoryStream and that errors...
It looks like this intersects with @clehner's recent changes to error handling
https://github.com/ssbc/scuttlebot/blob/914d63f60ab342cedc1338cfc3c8f4d1a90ddcb2/plugins/replicate.js#L216-L238

staltz:
By the way, I also quite often see the "wrong number" error. Scuttlebot v9.2.4

dominictarr:
@staltz are you replicating other devices locally? you should only see "wrong number" when answering the call (as the server) https://github.com/auditdrivencrypto/secret-handshake/blob/master/protocol.js#L92

pretty sure that error is getting dumped out here: https://github.com/dominictarr/multiserver/blob/master/compose.js#L65

someone is trying to connect to your pub/peer with a wrong key, but to get "wrong number" they have to get past the first stage, so they are using the same shs app key as you.

mmckegg:
@dominictarr this is how we came across the "fix":

- Mikey was running Patchwork(next) on the local area network with @mixmix and @pietgeursen who were both running standard sbot.
- I noticed that Mikey was continuously disconnecting and reconnecting from both mix and piet's sbot. This was chewing through quite a lot of bandwidth.
- I then fired up my patchwork. We noticed that Mikey and me had a solid connection (with no disconnects) but I was also continuously reconnecting to the others. 
- The question was: what is different about patchwork (over standard sbot)? 
  - Patchwork runs a forked `sbot/plugins/replicate.js` to add hooks for per peer progress. 
  - We discovered [this commit](https://github.com/ssbc/patchwork/commit/0b9adb1dfcfd2248f55f9dc2130f2d53aeb91794) which was something I changed before finally discovering that [this (fix realtime replication push #376)](https://github.com/ssbc/scuttlebot/pull/376) had been the actual problem. But I had left the other fix in patchwork since the old code looked pretty brittle to me (it would be running on pubs and not taking into account persistent gossip's prepopulated peers table).
- so we applied this fix to standard sbot, and the reconnecting problem seemed to go away (at least initially).

That's it really.

mmckegg:
@ahdinosaur did we test local peers without any patchwork in the mix (to see if they still get in reconnect loops)? Is it possible that this problem is actually caused by patchwork!?

dominictarr:
@mmckegg thanks for the context!
one question: do you have scuttlebot@9.4.6 in there?
(which has this: https://github.com/ssbc/scuttlebot/pull/374)
what scuttlebot version did @pietgeursen and @mixmix have?

staltz:
> someone is trying to connect to your pub/peer with a wrong key, but to get "wrong number" they have to get past the first stage, so they are using the same shs app key as you.

It could be that I was testing a couple of test accounts and they may have added my "real" account as a gossip peer and then the real account added those ones back as a gossip peer and maybe it's trying to connect to itself. Not sure, just speculations.

> are you replicating other devices locally?

I don't understand this question...

dominictarr:
> are you replicating other devices locally?
sorry, I meant do you have sbot running on more than once device - actually, in the video you posted I saw you using 2 laptops, I guess they where on the same network and scuttling?

staltz:
yeah, but even now that I'm not doing that, I get the error occasionally

dominictarr:
Okay, well that does not fit with my mental model of whats going on, so maybe my mental model is wrong.
Next time you get one, can you post the error? hmm if we improve the error logging in a little bit, we can add the ip address that we got the `wrong number` from - maybe you where scuttling but didn't realize.

dominictarr:
merged into 9.4.7

I'm gonna merge this and then open issues for the other things that have come up in this thread.

mixmix:
woop. I'm gonna pull this down and give it a whirl as a global install

dominictarr:
@mixmix can you tell me what version you had running before this? I want to know if you had @clehner's latest PR in there

mixmix:
I've been on 9.4.6 for a couple weeks. That's what  Iwas running at art
hack when mikey and matt were debugging

dominictarr:
oh hang on, this isn't publishing because `test/random.js` won't terminate

dominictarr:
hmm, no it must be caused by a dep because it's happening on older versions too

dominictarr:
yeah, sorry this isn't published it's not passing test/random.js @ahdinosaur @mmckegg  does it work for you?

dominictarr:
okay, i figured it out. it was actually a bug in my secret-handshake refactor. not in the crypto, just in the protocol glue! fixing it now...

dominictarr:
okay, this is fixed and published now. upgrade, and let me know how it goes.

mmckegg:
@dominictarr how do you feel about merging #381, now that this is sorted?

dominictarr:
@mmckegg done

PR #381, opened 2017-03-27
port patchwork style progress to scuttlebot

mmckegg:
As per ssbc/patchwork#447, this PR removes patchwork's need to be running a forked sbot version.

## Compatibility

I was going to remove the old `progress` but it turns out some tests need it to pass, so I left it in. I have also left the old `total` value in. This is kind of broken because `total - progress` usually never quite makes it to `0`, but it is better than it used to be (there was an off by 1 error in the old code).

The old `feeds` was broken since adding persistent gossip. From the code, it looked like it was supposed to count the number of feeds currently syncing, but since they never stopped, it didn't decrement until the peer disconnected. Seemed a little silly to me. **So I've changed it to now be the total number of feeds we know about**.

### `incompleteFeeds`

It gives us the total number of feeds that connected peers _claim_ they have more recent than us. This avoids the problem of the old system that sometimes never made it to 0. If we disconnect from a peer and didn't reach the _claimed_ max sequence, we stop counting it as a pending feed. 

Used for the **downloading messages** progress state in patchwork. 

### `pendingPeers`

It gives us the total number of pending feeds for each peer still syncing.

```
{ '@uRECWB4KIeKoNMis2UYWyB2aQPvWmS3OePQvBj2zClg=.ed25519': 1495,
   '@7nTbqZVRYRo7FokY8BBYGpXvvveRPb4u+G1vunRmNn4=.ed25519': 1425,
   '@J2VKbbf69rK38vXIRLOCwaZdEm+0vzv/LMeaxmJks3k=.ed25519': 862 
}
```

Used for the sync status progress spinners under "Connected Pubs" in the patchwork sidebar.

### add `sync` event on `createFriendStream`

Otherwise we don't know (when running with `live: true`) that we have finished loading the friends list. Used for only broadcasting `feeds` once we actually know how many feeds there are.

There is possibly room for some discussion here, but, eh. Works for patchwork.

### DIFF
--- a/plugins/replicate.js
+++ b/plugins/replicate.js
@@ -1,11 +1,13 @@
 'use strict'
 var pull = require('pull-stream')
+var pullNext = require('pull-next')
 var para = require('pull-paramap')
 var Notify = require('pull-notify')
 var Cat = require('pull-cat')
 var Debounce = require('observ-debounce')
 var mdm = require('mdmanifest')
 var apidoc = require('../lib/apidocs').replicate
+var deepEqual = require('deep-equal')
 
 var Pushable = require('pull-pushable')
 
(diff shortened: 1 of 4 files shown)


plugins/replicate.js
```diff
@@ -28,7 +24,7 @@ var streamErrors = {
   'read ETIMEDOUT': true,
   'write ECONNRESET': true,
   'write EPIPE': true,
-  'stream is closed': true, // rpc method called after stream ended
+  'stream is closed': true // rpc method called after stream ended
```
clehner: fwiw, the trailing comma was deliberate, so that future commits adding or removing lines could have cleaner diffs (as they would not have to add a comma to the last line)

mmckegg: Whoops, sorry! Call me a "standard" zealot, I can't even tell when I'm correcting things 😆

dominictarr:
okay, looks good. not gonna grumble about the weird es6 stuff. I'm working on a new better replication protocol, but I think we can make a progress thing that is backwards compatible with this.

dominictarr:
merged into 9.5.0, thanks @mmckegg 🎉

mmckegg:
Awesome! Whoops, I've got so used to dropping the keys now I forgot it was es6. One of my (only) favourite features :)

dominictarr:
dang whippersnappers, not even writing proper javascript anymore

PR #383, opened 2017-04-04
send sync over history stream

dominictarr:
this avoids making two requests to createHistoryStream (which doubles bandwidth used by the handshake, which for 1k feeds is a lot!)

The problem with this method, is that it depends on the bug in createHistoryStream https://github.com/ssbc/secure-scuttlebutt/issues/163

but the problem with this is it depends on what the remote sends us back, and it's not known if they have this bug or not. we could fix it but then we couldn't know for sure whether it's rolled out to every peer.

This is used for progress bars though, and the previous way was to compare with the sequence they have requested for this feed - are there any problems with that?

thoughts @mmckegg @clehner ?

### DIFF
--- a/plugins/block.js
+++ b/plugins/block.js
@@ -48,6 +48,7 @@ exports.init = function (sbot) {
         //the recipient.
         pull.take(function (msg) {
           //handle when createHistoryStream is called with keys: true
+          if(msg.sync) return true
           if(!msg.content && msg.value.content)
             msg = msg.value
           if(msg.content.type !== 'contact') return true
--- a/plugins/replicate.js
+++ b/plugins/replicate.js
@@ -312,29 +312,22 @@ function createHistoryStreamWithSync (rpc, upto, onSync) {
   // know when it switches to live. Do it manually!
   var last = (upto.sequence || upto.seq || 0)
   var state = null
-  return pullNext(function () {
-    if (!state) {
-      state = 'old'
-      return pull(
-        rpc.createHistoryStream({
-          id: upto.id,
-          seq: last + 1,
-          live: false,
-          keys: false
-        }),
-        pull.through(msg => {
-          last = Math.max(last, msg.sequence)
-        })
-      )
-    } else if (state === 'old') {
-      state = 'sync'
-      onSync && onSync(true)
-      return rpc.createHistoryStream({
-        id: upto.id,
-        seq: last + 1,
-        live: true,
-        keys: false
-      })
-    }
-  })
+  return pull(
+    rpc.createHistoryStream({
+      id: upto.id,
+      seq: last + 1,
+      sync: false,
+      live: true,
+      keys: false
+    }),
+    pull.filter(msg => {
+      if(msg.sync) return false
+      last = Math.max(last, msg.sequence)
+      return true
+    })
+  )
 }
+
+
+
+
--- a/plugins/replicate.js
+++ b/plugins/replicate.js
@@ -320,9 +320,12 @@ function createHistoryStreamWithSync (rpc, upto, onSync) {
       keys: false
     }),
     pull.filter(msg => {
-      if (msg.sync) return false
-      onSync && onSync()
-      return true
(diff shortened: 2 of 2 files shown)


mmckegg:
😱 ... ha! Well it works! And is much better than my crappy hack in terms of efficiency.

The real question is how do we fix https://github.com/ssbc/secure-scuttlebutt/issues/163 going forward without breaking this again!?

dominictarr:
the problem is if we merge _both_ of these then they break each other!
I'd rather fix this by switching to https://github.com/ssbc/ssb-ebt
so I'm gonna look into that first, maybe it won't be that complicated.

mmckegg:
@dominictarr I've spent some time considering this. As nasty as it feels, I think we should just merge it! And document as nasty legacy behaviour somewhere.

Lets not merge [the other fix](https://github.com/ssbc/secure-scuttlebutt/issues/163), or maybe we could remove sync check altogether and always send sync.

And you're right that this goes away with EBT, so whatever sinful decisions we make today will be absolved later!

dominictarr:
this seems like the plot of a greek tragedy, an api where false means true is like, also I'm not sure what other pubs use what versions. hmm, we should have made a thing where pubs post a message about what version they are running.

ahdinosaur:
> hmm, we should have made a thing where pubs post a message about what version they are running.

that'd have to be an ephemeral message (not on the log), unless it posts on every upgrade.

ahdinosaur:
by the way, keen to merge this, but can we add documentation somewhere that `sync` is false means true?

mmckegg:
Okay lets merge this and #386 then release a new version! Anything else to roll in?

I'll do it tonight.

All **not** in favour please say "nooooo!!!!!" 😼

mmckegg:
@dominictarr I just tested this, and it doesn't actually work. Even with the `sync: false` hack, the `sync` messages never arrive 😢

mmckegg:
I tested this and unfortunately it doesn't actual work. The `sync` event is never emitted even with the hack.

😢

ahdinosaur:
@mmckegg what is happening here: https://github.com/pull-stream/pull-live/blob/master/index.js#L18?

ahdinosaur:
bread crumbs: [createHistoryStream](https://github.com/ssbc/secure-scuttlebutt/blob/master/indexes/clock.js) -> [Follower](https://github.com/ssbc/secure-scuttlebutt/blob/master/follower.js) -> [pull-level.read](https://github.com/pull-stream/pull-level/blob/master/read.js) -> [pull-live](https://github.com/pull-stream/pull-live/blob/master/index.js)

mmckegg:
@ahdinosaur That's just emitting a `{sync: true}` item if `opts.sync` is not `false`.

ahdinosaur:
@mmckegg hm, i'm trying to debug this now, but i'm getting a sync message. :open_mouth: 

- cloned `scuttlebot` and `secure-scuttlebutt`, `npm link`-ed them, and checked out this branch
- i changed [this line](https://github.com/ssbc/secure-scuttlebutt/blob/master/indexes/clock.js#L32) to be `sync: true`
- ran `./bin.js server` (or else got to make sure to run `noderify bin.js > sbot.js` to rebuild `sbot.js`)
- ran `sbot createHistoryStream --id '@6ilZq3kN0F+dXFHAPjAwMm87JEb/VdB+LC9eIMW3sa0=.ed25519' --live`

dominictarr:
what if we compared sequence to the createHistoryStream call we received, and if we didn't get that, then just consider this stream synced once it becomes quiet for a timeout (say, 3 seconds)

mmckegg:
@dominictarr that could work. Was thinking about that while sleeping a few nights back.

It's just `progress`, it really doesn't deserve the right to double bandwidth use.

mmckegg:
@dominictarr I'll have a go at [this](https://github.com/ssbc/scuttlebot/pull/383#issuecomment-292918680) later today!

PR #385, opened 2017-04-07
Allow setting a note for an invite code

clehner:
%UcqW5DcDjD8U9PbYxTnCz9Zl+7HoUTf8usOnRA4uuOA=.sha256

### DIFF
--- a/plugins/invite.js
+++ b/plugins/invite.js
@@ -60,11 +60,15 @@ module.exports = {
     })
 
     return {
-      create: valid.async(function (n, cb) {
+      create: valid.async(function (n, note, cb) {
         var modern = false
         if(isObject(n) && n.modern) {
           n = 1
           modern = true
+        }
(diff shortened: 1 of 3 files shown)


dominictarr:
sorry I fogot to merge this, just looked at the code, and it should use just one argument (not counting cb) because with two args you can't use from the cli. one argument is easy to translate from optimist options, so the note should be an option `sbot.create invite --uses 5 --note "invite for bob"` for example.

clehner:
two arguments works from the CLI. e.g.

sbot invite.create 50 https://github.com/ssbc/scuttlebot/wiki/Pub-Servers

named arguments are probably better for usability and extensibility though, so i could change this

dominictarr:
I merged, but changed to `--uses N --note "this is bob"`

dominictarr:
merged into 9.8.0

clehner:
thanks!

PR #386, opened 2017-04-07
fix progress stalling when pubs disconnect before completion

mmckegg:
Just noticed that sometimes (particularly right now with the high load on pubs) the "downloading messages" progress stalls and never completes. The only way to make the progress bar go away is to restart.

Turns out it is a display only issue. This PR correctly handles cleanup when a pub disconnects before sync completes.

### DIFF
--- a/plugins/replicate.js
+++ b/plugins/replicate.js
@@ -288,7 +288,10 @@ module.exports = {
                 }
               }
 
-              pendingFeedsForPeer[rpc.id].delete(upto.id)
+              if (pendingFeedsForPeer[rpc.id]) {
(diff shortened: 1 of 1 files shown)


mmckegg:
@dominictarr @ahdinosaur anyone mind if I just merge this? That stuck progress is really annoying 😢 and I'm trying to resist forking again! 😆 

Also not sure if I have the npm publish rights.

ahdinosaur:
:+1:

dominictarr:
go ahead!

ahdinosaur:
@dominictarr can you add @mmckegg to be an npm owner?

dominictarr:
done

PR #393, opened 2017-04-11
replicate: reduce bandwidth by replacing double stream with peerHas check

mmckegg:
Continuing from #383. Here's an alternative strategy that uses what we know about the peer already to decide when to mark as sync. It also falls back to a timeout after 3 seconds of no message received. 

Does this look okay? @dominictarr @clehner @ahdinosaur

### DIFF
--- a/plugins/replicate.js
+++ b/plugins/replicate.js
@@ -336,7 +336,6 @@ function detectSync (peerId, upto, toSend, peerHas, onSync) {
     } else {
       // if we get here, the peer hasn't yet asked for this feed, or is not responding
       // we can assume it doesn't have the feed, so lets call sync
-      console.log('timeout')
       broadcastSync()
     }
   }, 500)
--- a/plugins/replicate.js
+++ b/plugins/replicate.js
@@ -273,7 +273,7 @@ module.exports = {
             }),
 
             // track sync completed progress
-            pull.through(detectSync(rpc.id, upto, peerHas, function () {
+            pull.through(detectSync(rpc.id, upto, toSend, peerHas, function () {
               if (pendingFeedsForPeer[rpc.id]) {
                 pendingFeedsForPeer[rpc.id].delete(upto.id)
                 debounce.set()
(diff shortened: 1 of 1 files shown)


clehner:
I haven't deeply evaluated it but the idea sounds good

dominictarr:
@mmckegg can you explain how it works? any special things we should notice? this is the heart of ssb so I'm not just gonna LGTM and merge

mmckegg:
Sure, basically it does "nothing" at all.

That is, it **shouldn't** touch any of the output. Things should be exactly how they used to be before I [added that dodgy sync hack for `progress`](https://github.com/ssbc/scuttlebot/pull/381).

All this does is adds a hook as a through stream and watches for when the server is up to date with what it claims and then emits an event that is consumed by the progress stuff to signal that a feed has finished syncing.

I realise now looking at it that it is very hard to read and see the logic flowing. I'll try restructuring it a bit to make it clearer what's actually going on.

dominictarr:
thank you @mmckegg a bit of context makes it much much easier to know what to look for when code reading. it can be hard to figure out things like intention just looking at code.

dominictarr:
Okay, so one thing about this design is that it should work fine on initial sync, but if you don't actually receive any messages, you don't know what the progress is, but then you fallback to the time out.
Question: what happens in these situations

1. you (alice) have a low seq, but the other (bob) has high. -> you'll make that feed as synced when you receive the last message.
2. alice has high, but bob has low -> not tracked, this is just hit by the timeout.
3. alice requests a feed, but bob doesn't have that feed at all. -> I think this still creates a live feed (incase bob does decide to follow) so this is also handled by a time out.
4. alice doesn't request a feed but bob does. -> this isn't represented in the progress.
5. alice and bob both request the same feed but are already in sync -> handled by timeout?

> note, in the above, 2,4 are the compliment of 1,3. when alice is in situation 1, bob is in 2. when alice is in 3, bob is in 4, and vice versa. 5 is the special case. It's a bit mind bending because you have to visualise two peers which may be reflections of each other or swap positions. after the -> is what _alice_ does. (bob does the thing that alice would do in the complementary situation)

@mmckegg does that look right? I think 3 (alice follows something bob doesn't ask for is the only one that really needs a time out?)

mmckegg:
@dominictarr thanks for that analysis! I think I've managed to address all of those points now.

I've changed the way timeout works so instead of potentially timing out after any message, it waits 500ms for the peer to request a particular feed, and if it hasn't requested it by then, assumes they don't have it and calls sync.

If they have requested a feed, check to see if it is newer than the sequence we have. If it is, wait for them to send us the new messages, otherwise call sync immediately.

We are completely ignoring sending stuff, only receiving. Great for initial sync, but also the rest of the time I don't think it is all that useful to know that you're still sending stuff out. Eh... this is fine for now.

mmckegg:
Can we just merge this? It is reverting back to old scuttlebot behaviour (saving tonnes of bandwidth), and no one cares if progress is a little screwy (which I'm pretty sure is fine now).

@dominictarr @clehner

dominictarr:
oh yeah, sorry i forgot, this is at least as good as it was before, so am merging.

dominictarr:
merged into 9.6.1 !

PR #400, opened 2017-04-20
enable configuring no gossip, or connect to seeds only

dominictarr:
@pietgeursen needed this, so I cherry-picked it from the flume branch.

### DIFF
--- a/plugins/gossip/index.js
+++ b/plugins/gossip/index.js
@@ -81,6 +81,7 @@ module.exports = {
 
         if(!addr.key) return cb(new Error('address must have ed25519 key'))
         // add peer to the table, incase it isn't already.
+        console.log('connect', addr)
         gossip.add(addr, 'manual')
         var p = gossip.get(addr)
         if(!p) return cb()
@@ -264,3 +265,5 @@ module.exports = {
     return gossip
   }
 }
+
+
--- a/plugins/gossip/schedule.js
+++ b/plugins/gossip/schedule.js
@@ -44,7 +44,12 @@ var isOnline = not(isOffline)
 function isLocal (e) {
   // don't rely on private ip address, because
   // cjdns creates fake private ip addresses.
-  return ip.isPrivate(e.host) && e.source === 'local'
+  // ignore localhost addresses, because sometimes they get broadcast.
+  return !ip.isLoopback(e.host) && ip.isPrivate(e.host) && e.source === 'local'
+}
+
+function isSeed (e) {
+  return e.source === 'seed'
 }
 
 function isFriend (e) {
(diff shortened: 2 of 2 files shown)


dominictarr:
merged into 9.7.0, 
oh btw, this makes it possible to actually turn off different types of gossip.

``` js
"gossip": {
  "local": false, //do not replicate peers on the local network
  "global": false, //do not replicate with peers on the internet
  "seeds": false, //do not replicate with seeds.
},
//seeds to connect to my pub, and cel's pub.
"seeds": [
    "net:104.236.22.57:8008~shs:5XaVcAJ5DklwuuIkjGz4lwm2rOnMHHovhNg7BFFnyJ8=",
    "net:128.199.132.182:8008~shs:DTNmX+4SjsgZ7xyDh5xxmNtFqa6pWi5Qtw7cE8aR9TQ="
  ],
```

seeds are optional, but have been useful for developing ebt.

arj03:
Nice, would be good to add that comment to the docs.

kseistrup:
Is there a way to ban sbot from connecting to certain hostnames and/or IP addresses?

I run sbot with `--allowPrivate` because I want to include the fc00::/8 address space, and it seems sbot picks up, e.g., `localhost:9000` via gossip. Can I tell sbot to ignore that without being overly broad?

dominictarr:
@kseistrup right so you want to connect to cjdns but not the regular internet? localhost:9000 is really a separate problem - there are some badly formed pub announcements which say localhost, and they should just be ignored

kseistrup:
@dominictarr No, I want sbot to connect to the regular internet (IPv4 and IPv4) _and_ any fc00::/8 addresses it might have heard of, and I want it to ignore localhost:9000 (mostly because it spams my syslog, but also because it's a waste of time).

dominictarr:
got it. hmm, probably it should just ignore localhost unless it's provided as a seed (this is used in the tests)
This isn't a configuration problem though, it's a code problem.

kseistrup:
I agree about ignoring localhost.

Still, couldn't it be useful to have a way to ban/mute/keelhaul individual nodes?

dominictarr:
@kseistrup sure, I would gladly merge PRs for either or both!

kseistrup:
I'd be more than happy to. Alas, I don't speak js.

dominictarr:
@kseistrup the changes you are talking about wouldn't be very complicated, if you can do any programming language you could probably figure it out, we'd be happy to help guide you, but given that there is a huge bredth of things that need doing, trying and asking for help would make it happen a lot faster than waiting for someone else to get around to it.

kseistrup:
@dominictarr Thanks for the advice and optimism, I'll see what I can do.

PR #401, opened 2017-04-23
remove (unintended?) console.log from gossip

mmckegg:
https://github.com/ssbc/scuttlebot/commit/ac51d55eb952fb66353ffa811b94522fee651e1a#commitcomment-21850365

Probably not supposed to be here, right?


dominictarr:
thanks

PR #403, opened 2017-05-19
Use ssb-ref 2.7.1 which works for IPv4 and IPv6

staltz:
Current scuttlebot depends on ssb-ref `^2.6.2`, which for some people due to semver might be installed as 2.6.**3**. Version 2.6.3 is pooped because it [broke support for IPv4](https://github.com/ssbc/ssb-ref/issues/10). I suspect some people might be getting problems with this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "ssb-config": "^2.0.0",
     "ssb-keys": "^7.0.0",
     "ssb-msgs": "~5.2.0",
-    "ssb-ref": "^2.6.2",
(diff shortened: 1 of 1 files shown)


staltz:
Hmm, as usual I got confused around the `^` notation. It means `^2.6.2` will fetch `2.7.1`, so maybe it's not a problem, but still would be good to update this nevertheless.

dominictarr:
thanks, merging

dominictarr:
merged into 9.8.0!

PR #407, opened 2017-05-31
merge all the flume stuff! (also add blocking)

dominictarr:
This is all ready now, y'all!

this merges all the flume stuff, as well as adds blocking. The replication policy is all shifted out into ssb-friends, so that is easy to change. It's also possible to replicate someone non persistently, so you could replicate someone when you open their profile, but without following them - their feed would not be replicated next time you start ssb.

to block someone:

``` js
sbot publish --type contact --contact <feed_id> --blocking
```

I just gotta update ssb-ebt to some of the new ideas in here, but this is all ready to merge!

**important**: when you start this after using old scuttlebot it will migrate from the old database to a flume one, which will take a while. if you want to check the progress, use `sbot.since` and that will dump the state of all the views, when the numbers stop changing, your are ready.

PR #408, opened 2017-06-02
Improve robustness of loading user-plugins

clehner:
- Skip loading a plugin with an already-used name,
  as mentioned in https://github.com/ssbc/scuttlebot/pull/407#issuecomment-305742813
- Add back plugin loading error checking removed in 5653a26d6cfcd9a9cff414d2c291fc24e7a294aa

This addresses https://github.com/ssbc/scuttlebot/pull/407#issuecomment-305731499 so that a user upgrading to the flume branch will not get crashes if they have previously configured user-plugins which are now loaded by scuttlebot core. (these being ssb-ws, ssb-query and ssb-links) - instead, they will get a one-line error message for each, saying that the plugin is already loaded.

I also reenabled wrapping loading the plugins in a try-catch block and checking that the plugin object is valid with assertSbotPlugin, for robustness.

### DIFF
--- a/plugins/plugins.js
+++ b/plugins/plugins.js
@@ -175,8 +175,17 @@ module.exports.loadUserPlugins = function (createSbot, config) {
   //instead of testing all plugins, only load things explicitly
   //enabled in the config
   for(var k in config.plugins) {
-    if(config.plugins[k])
-      createSbot.use(require(path.join(nodeModulesPath, k)))
+    if(config.plugins[k]) {
+      try {
+        var plugin = require(path.join(nodeModulesPath, k))
+        assertSbotPlugin(plugin)
+        if (createSbot.plugins.some(plug => plug.name === plugin.name))
+          throw new Error('already loaded')
+        createSbot.use(plugin)
+      } catch (e) {
+        console.error('Error loading plugin "'+k+'":', e.message)
+      }
+    }
   }
 }
 


dominictarr:
this looks good, but I think it should exit on the error. usually if you don't have a plugin you expected then it won't work right.

dominictarr:
merged but made it throw instead.

dominictarr:
Oh I thought that this is fixed by explain-error@1.0.4, but maybe not (it might change the name?) merged your PR to flumeview-level, is there anywhere else?

clehner:
On Friday, June 2, 2017 1:21:33 PM HST, Dominic Tarr wrote:
> Oh I thought that this is fixed by explain-error@1.0.4, but 
> maybe not (it might change the name?) merged your PR to 
> flumeview-level, is there anywhere else?

for the NotFoundError? a quick check shows that explain-error passes 
through error.name now, which is great - so flumedb/flumeview-level#1 is 
not necessary now, but shouldn't hurt.

clehner:
> +    if (createSbot.plugins.some(plug => plug.name === k))

this does not work, because k may be `ssb-ws`, but plug.name is `ws`. that 
is why i was loading the plugin before checking the name.

clehner:
@dominictarr i just want to point out 
https://github.com/ssbc/scuttlebot/pull/408#issuecomment-305931508 again, 
that commit 95671f8a1c271f96a96d605f9cc83598e0a7f337 is problematic. i do 
this only to try to help make it easier for people when they update

dominictarr:
@clehner oh sorry. yes you are correct. though, I think this behaviour will lead to surprises and edge cases down the road. Could we just enforce that plugin names are just module names with the `"ssb-"` prefix removed? I think this would work for most plugins (except ssb-links which creates links2) but we can just republish that as ssb-links2?

clehner:
patchfoo and git-ssb-web as plugins don't begin with "ssb-". also sometimes 
i have multiple versions of a plugin in my ~/.ssb/node_modules, like 
git-ssb-web-git, git-ssb-web-npm, where i enable one at a time, and the 
exported plugin.name does not necessarly correspond to the directory name. 
but that could just be me that does that. other than that, i think what you 
said would work

dominictarr:
hmm, okay what if cases where a name changed was totally explicit?

``` js
plugins: {
  <name>: <module_name>,
  patchfoo: true //or just true if name===module_name,
  links2: ssb-links
}
```

it would require changing people's configuration, but it's better to clean it up now while it effects only a few people. if there is something that doesn't make sense, like `links: true` that tries to create a module named links2 then we'll error with a helpful message.

clehner:
ah, i like that, since it maps onto what is loaded on the rpc object. and 
could allow for loading from arbitrary paths for development, by the looks 
of it

dominictarr:
yes! we could make it just be a path.

PR #411, opened 2017-06-09
implement plugin name mapping

dominictarr:
as discussed at https://github.com/ssbc/scuttlebot/pull/408#issuecomment-306954532

except the other way around:
```
"plugins": {
  <module_name>: <plugin_name>, //if plugin name is not the same as plugin name
  ssb-<plugin_name>: true, //if plugin name is the same as module name, but with ssb- prefix.
  plugin_name: true //if plugin name and module name are the same.
}
```

an error will be thrown if the plugin doesn't have the expected name, for example, ssb-links creates plugin named links2 so that will error.

I didn't implement a thing to automatically set up the plugins name if it's different. to be honest just use npm install directly.


arj03:
I'm getting this error:

Error: plugin at:ssb-links expected name:links but had:links2
    at Object.module.exports.loadUserPlugins (/home/chrx/dev/scuttlebot/plugins/plugins.js:187:15)

PR #412, opened 2017-06-09
merge latest legacy replication into flume from master

mmckegg:
From https://github.com/ssbc/scuttlebot/pull/407#issuecomment-307357933:

> I just noticed that we are not using the latest legacy replication code. It's fallen back to the old one before we fixed the double pull stream hack that I did for progress sync back in the day.
>
> We should be using this: https://github.com/ssbc/scuttlebot/blob/ea4748e7991a1f992b96faf319cbfd5059639a03/plugins/replicate.js

### DIFF
--- a/lib/detect-sync.js
+++ b/lib/detect-sync.js
@@ -1,0 +1,48 @@
+module.exports = function detectSync (peerId, upto, toSend, peerHas, onSync) {
+  // HACK: createHistoryStream does not emit sync event, so we don't
+  // know when it switches to live. Do it manually!
+
+  var sync = false
+  var last = (upto.sequence || upto.seq || 0)
+
+  // check sync after 500ms, hopefully we have the info from the peer by then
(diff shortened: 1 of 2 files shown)


dominictarr:
hmm, sometimes git gets confused when you are merging between branches and doesn't merge a thing as you intended. wondering if smaller files are less confusing? because then you'll have something more like k:v which is very easy to merge, not lists which are somewhat less simple?

mmckegg:
@dominictarr I'm not sure what you are asking. Is there something you want me to do? 

Is GitHub showing you something different than I'm seeing?

dominictarr:
sorry, I was just commenting. merging this.

dominictarr:
oh this is breaking one test: test/block3.js

dominictarr:
oh but it's an easy fix

dominictarr:
merged into 10.0.3

PR #414, opened 2017-06-09
Fix npm bin-script path

kemitchell:
After installing 10.0.0, the bin-script went empty on me:

```shellsession
npm i -g scuttlebot
cat $(which sbot)
```

Turns out the `bin` script path in `package.json` is off.  From the looks of it, `sbot.js` was renamed to `cli.js` without a corresponding `package.json` change?  I've done that more times than I care to count.

Here is the corresponding `package.json` change.

Thanks to all!

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "tape": "~4.0.0"
   },
   "bin": {
-    "sbot": "./sbot.js"
+    "sbot": "./bin.js"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


kemitchell:
Oh, bummer. Didn't see #371.

dominictarr:
hmm, I think we don't really need the bundled sbot.js anymore.

PR #415, opened 2017-06-10
remove old `loadedFriends` check that doesn't work with flume

mmckegg:
This PR removes old `loadedFriends` check that doesn't work with flume and breaks feed sync progress

This is currently stopping sync progress from working in Pathwork


dominictarr:
great, oh, we need to make replication show on the new progress api {start,current,target} format. I just added a progress bar to sbot so it shows if you are migrating or rebuilding indexes.

dominictarr:
merged into 10.0.4

PR #424, opened 2017-06-30
gossip: debounce has-network for better windows performance

mmckegg:
I have been trying to figure out why Patchwork runs so badly on my girlfriend's computer. It's an old thinkpad from 2011 running Windows 8. Finally hit a breakthrough in the profiler:

windows-misery

Each connection attempt is blocking for more than 2 seconds!! Tracked this down to many calls to `hasNetwork`. It was being called 100s of times a second. And apparently it is a slow blocking command on this machine. Haven't investigated why it is so slow, instead I fixed it the quick and dirty way.

**This PR adds a 1 second debounce to all `hasNetwork` requests. And suddenly everything  runs beautifully on that machine.**

cc @dominictarr

### DIFF
--- a/lib/has-network-debounced.js
+++ b/lib/has-network-debounced.js
@@ -1,0 +1,12 @@
+var hasNetwork = require('has-network')
+
+var lastCheck = 0
+var lastValue = null
+
+module.exports = function hasNetworkDebounced () {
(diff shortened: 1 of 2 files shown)


dominictarr:
oops! merging

dominictarr:
merged into 10.3.1

PR #425, opened 2017-06-30
fix db.ready check for local peers

mmckegg:
Noticed that while adding the `db.ready` check on gossip scheduling solved most of the _replicate while migrating_ bugs, other machines can still connect to you, and you'll try to replicate with them.

**This PR adds some more `db.ready` checks on replication. Will immediately disconnect from a peer if it tries to connect in this time.** It also adds a missing `notify()` case for removed local peers.

Issue in patchwork: ssbc/patchwork#567

cc @dominictarr

### DIFF
--- a/plugins/gossip/index.js
+++ b/plugins/gossip/index.js
@@ -276,6 +276,10 @@ module.exports = {
     //get current state
 
     server.on('rpc:connect', function (rpc, isClient) {
+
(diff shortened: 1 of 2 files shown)


dominictarr:
looks good.

dominictarr:
thanks, merged into 10.4.0

PR #431, opened 2017-09-10
remove blocking hasNetwork

evbogue:
Something has been seriously wrong with scuttlebot, and my instinct was telling me that it had something to do with gossip. After a conversation with a Benjamin Franklin statue about non-blocking I/O in JS, I decided to poke around a little bit in the code. I discovered `hasNetworkdebounced`, which was new to me. I tracked back the pull, and found https://github.com/ssbc/scuttlebot/pull/424 -- which pointed to a _serious problem_, which is the whole application is blocking every time it checks to see if the Internet is up.

It occurred to me that checking to see if the Internet is up is I/O, which is slow and thus could block Node.js. After this realization, I delete-key-programmed a few lines of code, and now scuttlebot is working a lot better. I can even gossip with my pub now!

I think we should merge this until we figure out how to check and see if the connection is up without blocking the entire app in the process.

### DIFF
--- a/plugins/gossip/schedule.js
+++ b/plugins/gossip/schedule.js
@@ -2,7 +2,6 @@
 var ip = require('ip')
 var onWakeup = require('on-wakeup')
 var onNetwork = require('on-change-network')
-var hasNetwork = require('../../lib/has-network-debounced')
 
 var pull = require('pull-stream')
 
@@ -30,17 +29,6 @@ function maxStateChange (M, e) {
 function peerNext(peer, opts) {
   return (peer.stateChange|0) + delay(peer.failure|0, opts.factor, opts.max)
 }
-
-
-//detect if not connected to wifi or other network
-//(i.e. if there is only localhost)
-
-function isOffline (e) {
-  if(ip.isLoopback(e.host) || e.host == 'localhost') return false
-  return !hasNetwork()
-}
-
-var isOnline = not(isOffline)
 
 function isLocal (e) {
   // don't rely on private ip address, because
@@ -130,7 +118,7 @@ function (gossip, config, server) {
     }
 
     //will return [] if the quota is full
-    var selected = select(peers, ts, and(filter, isOnline), opts)
+    var selected = select(peers, ts, and(filter), opts)
     selected
       .forEach(function (peer) {
         gossip.connect(peer)


staltz:
Why does it still work, though? Is it because of this line that checks for connection to each peer (which would be impossible if hasNetwork == false)?

https://github.com/ssbc/scuttlebot/pull/431/files#diff-0dc8dd56b0edd526a37dd898541c44e5R84

evbogue:
@staltz I think the line you're pointing to checks to make sure scuttlebot doesn't have a current connection open with a peer, before it connects to it? But all of the double negatives make my head spin. 

What this pull does is eliminate the blocking that was happening every couple of seconds because checking to see if the Internet is up or down blocks Node.

I think the gossip schedule could use solid re-write. Maybe now is the time? I think the gossip schedule should: 

+ Build a list of peers (init.js does this quite well)
+ Connect to them
+ If they are down, ignore them for awhile

dominictarr:
Yeah I hate the gossip scheduling code...
but I want to understand _why_ this is a problem.

is it blocking _in_ hasNetwork, or is it blocking in hasNetwork's callback? for example, would this fix it?

```
diff --git a/plugins/gossip/schedule.js b/plugins/gossip/schedule.js
index 0a7b08f..1b0ec39 100644
--- a/plugins/gossip/schedule.js
+++ b/plugins/gossip/schedule.js
@@ -109,7 +109,7 @@ function (gossip, config, server) {
 
   //trigger hard reconnect after suspend or local network changes
   onWakeup(gossip.reconnect)
-  onNetwork(gossip.reconnect)
+//  onNetwork(gossip.reconnect)
 
   function conf(name, def) {
     if(config.gossip == null) return def
```
we can't just delete hasNetwork... because we need to know whether or not we are online... we just need a way to test that without blocking

dominictarr:
We need to measure how _much_ hasNetwork blocks, and in what circumstances maybe it only happens on some comptuers, or OS or something?

staltz:
It depends on `os.networkInterfaces()` which returns synchronously. It could be that in some operating systems that retrieval is slow. The alternative would be reimplement this Node.js core API using e.g. `exec` and calling `ifconfig`.

evbogue:
I think this might still be an issue, but I'm closing this pull request as it is not the correct solution to the problem.

PR #432, opened 2017-09-12
Remove hack and fix blobs.add on cli

regular:
Update ssb-client to 4.5.0 and removes the blobs.add hack (which is part of ssb-client since ssb-client@4.5.0)

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -8,7 +8,6 @@ var File         = require('pull-file')
 var explain      = require('explain-error')
 var ssbKeys      = require('ssb-keys')
 var stringify    = require('pull-stringify')
-var createHash   = require('multiblob/util').createHash
 var minimist     = require('minimist')
 var muxrpcli     = require('muxrpcli')
 var cmdAliases   = require('./lib/cli-cmd-aliases')
@@ -115,10 +114,6 @@ if (argv[0] == 'server') {
       cb()
     }
 
-    // HACK
-    // we need to output the hash of blobs that are added via blobs.add
-    // because muxrpc doesnt support the `sink` callback yet, we need this manual override
-    // -prf
     if (process.argv[2] === 'blobs.add') {
       var filename = process.argv[3]
       var source =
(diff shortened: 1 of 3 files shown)


regular:
Hmm ... this seems to make the tests hang forever

dominictarr:
thats a manifest problem. I think the manifest needs to be set on the client to include createHistoryStream

dominictarr:
btw, which test hangs? some of them do take a while, normally if it's test/random.js it's already, although test/blob* is pretty good at hanging too

regular:
"# test invite with note" is the last one that is started.
```
ok 17 should be equivalent
```
is the last assertion before it hangs.

christianbundy:
@regular Is this still something that you'd like to see merged?

regular:
@christianbundy on one hand, it's not good that we have workarounds in two separate places for the same issue. On the other hand, things evolved a bit since I opened this, so it would require some more work and I currently don't have the capacity to deal with this. I had lots of issues with blobs.add on the command line in the past, it would error every now and then and not add the blob (2/3 chance of success roughly) I guess that's related but haven't tracked it down.

dominictarr:
btw: it would be cool to fix this in muxrpc - so that sink streams could callback a value (would also mean a strong confirmation that the stream succeeded) I'd estimate the effort there as a day or two. but the only thing that needs that is `blobs.add` so it's a lot of work for little reward. Better to leave it until a new version of muxrpc. (I did try to write one last southern summer , and it was working, but deploying it screwed up everything because I hadn't implemented back pressure well (the infamous muxrpc@7)- iirc I did consider supporting this case though)

regular:
@dominictarr I am wondering: what if muxrpc would support transform streams instead? I have vague memories of wanting that for something in the past. `blobs.add` then could _transform_ the data stream into a hash and this pattern might be more useful in general than a sink with callback?

dominictarr:
@regular that's essentially how it would work internally. really muxrpc only has duplex stream, but it lets you turn off one side (except for errors) to make a source or sink stream. `blobs.add` would be a stream with just one chunk of data (the hash) coming back...

although, I do think it's important that the data coming back can mean the source has succeeded, which is what a cb is for...

hmm, idea: what if you implemented this entirely in user space? and exposed it as duplex? it doesn't seem that hard to do it that way. (hard part is wrangling all the muxrpc tests...) so maybe it isn't that hard.

christianbundy:
Is this still active and/or ready for review?

clehner:
Merged. Thanks @regular.

PR #436, opened 2017-09-29
prevent gossip scheduling while downloading messages

mmckegg:
The current PR contains a quick hack to prevent gossip from being scheduled while feed download is taking place. It detects this by checking to see if a message was downloaded in the last 500ms.

Currently it uses a nasty global whacked on the `sbot` to share the state out of legacy replication into gossip scheduling.

---

But the important thing is just how much this improves onboarding performance. It no longer crashes! Completes in a few minutes instead of mysterious indefinite locking.

See https://github.com/ssbc/patchwork/issues/631 for more context.

cc @dominictarr @clehner @ahdinosaur

### DIFF
--- a/plugins/gossip/schedule.js
+++ b/plugins/gossip/schedule.js
@@ -137,6 +137,13 @@ function (gossip, config, server) {
       })
   }
 
+  function currentlyDownloading () {
+    // don't schedule gossip if currently downloading messages
+    if (server.lastMessageAt && server.lastMessageAt > Date.now() - 500) {
+      console.log('skip gossip')
+      return true
+    }
+  }
 
   var connecting = false
   function connections () {
@@ -146,7 +153,7 @@ function (gossip, config, server) {
       connecting = false
 
       // don't attempt to connect while migration is running
-      if (!server.ready()) return
+      if (!server.ready() || currentlyDownloading()) return
 
       var ts = Date.now()
       var peers = gossip.peers()
@@ -247,7 +254,3 @@ exports.isLocal = isLocal
 exports.isFriend = isFriend
 exports.isConnectedOrConnecting = isConnect
 exports.select = select
-
-
-
-
--- a/plugins/replicate/legacy.js
+++ b/plugins/replicate/legacy.js
@@ -243,6 +243,8 @@ module.exports = function (sbot, notify, config) {
 
       debounce.set()
 
+      var sync = false
+
       pull(
         rpc.createHistoryStream({
           id: upto.id,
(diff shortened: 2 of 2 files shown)


mmckegg:
@ahdinosaur  For some reason I don't think @dominictarr will like this 😆

evbogue:
@mmckegg, do you mind if I ask what issue this is attempting to solve?

It's my opinion that the gossip schedule is a mess right now. Partially this is because we are between ssb-ebt and old replication, the other part is that it's hard to tell wtf is going on with the gossip schedule. 

I'll speculate that all current ssb bugs are related to this plugin.

mmckegg:
@evbogue I noticed that sbot starts connecting to pubs it learns about as it is still onboarding (e.g. from a local peer). It then starts downloading from all of these pubs at the same time. Requesting the same stuff from all the peers. This causes CPU and network usage to spiral out of control sometimes leading to crash, but the biggest issue is it causes lockups which makes the initial experience very slow and unresponsive. I’ve had reports of initial download taking more than 4 hours. 

I made this hack to see if the situation could be improved by simply not connecting to other pubs while data is being downloaded. In my testing so far, seems like it has. 

I’m about to deploy this in the new patchwork 1.7.0. Will be interesting to see if it improves the situation for other people too!

plugins/gossip/schedule.js
```diff
@@ -137,6 +137,13 @@ function (gossip, config, server) {
       })
   }
 
+  function currentlyDownloading () {
+    // don't schedule gossip if currently downloading messages
+    if (server.lastMessageAt && server.lastMessageAt > Date.now() - 500) {
+      console.log('skip gossip')
```
arj03: Shouldn't this be: server.emit('log:info' ?

Maybe the message should be "already downloading messages from another peer, skip connect"

arj03:
@mmckegg whats the status here? It's been included since patchwork 3.7 it seems. Keen on seeing another release and this looks good.

mmckegg:
@arj03 Yeah, that console.log shouldn’t be in there. 

Seems to be working well in patchwork.

dominictarr:
sorry this must have slipped down my inbox....
I would much rather merge a PR where the parts communicated with each other in a cleaner way... but I hate the gossip scheduling code anyway and want to rewrite it at some point (but also hate it so much that I don't want to rewrite it either...!)
merged into 10.4.11

mmckegg:
> I would much rather merge a PR where the parts communicated with each other in a cleaner way

Yep 😢 I agree, this is pretty nasty. A stop-gap hack at best!

PR #438, opened 2017-10-01
fix updating from local peers that you don't follow

### LINKED ISSUE #437: messages aren't being gossiped after initial sync with local peers when not followed
Discovering a new peer on the network will trigger a feed download, but only the first time. It doesn't seem to get any changes to the feed after that initial sync.

Something fishy is going on!

Reproduce by running a fresh sbot on one machine (and add a post) then another on the same local network. The initial posts seem to appear, but subsequent replies do not unless you explicitly follow. Normally this is obscured by pub FOAFS, but the issue becomes apparent on LANs when someone has no pubs in common with you.

---

I haven't yet tested this exhaustively, so it might be something else going on (maybe in Patchwork). Can anyone else reproduce?

mmckegg:
The `toSend[id] == null` is an old check that is unnecessary after adding the `e.source === 'local'`. It stops replication from requesting feeds from a local peer after restart if you have not followed them, which is really not what you want.

**This PR removes that broken check, and now local peers replicate their own feed as expected!**

Gonna roll this out in the new version of patchwork about to be released.

fixes #437

### DIFF
--- a/plugins/replicate/legacy.js
+++ b/plugins/replicate/legacy.js
@@ -196,7 +196,7 @@ module.exports = function (sbot, notify, config) {
   function localPeers () {
     if(!sbot.gossip) return
     sbot.gossip.peers().forEach(function (e) {
-      if (e.source === 'local' && toSend[e.key] == null)
+      if (e.source === 'local')
         request(e.key)
     })
   }


mmckegg:
Been running this in patchwork for the last couple of weeks. Seems to be working well!

We should merge this!

cc @dominictarr

staltz:
We should merge this!

dominictarr:
okay, merging

alanshaw:
:thumbsup: yes this fixed the issue I was seeing!

dominictarr:
merged into 10.4.9

PR #440, opened 2017-10-09
Allow "error" logging level to be used

amsross:
`conf.logging.level: 'error'` was not previously accessible because the declaration would fallback to `DEFAULT_LEVEL` where `['error', ...].index('error')` returns a falsey value.

### DIFF
--- a/plugins/logging.js
+++ b/plugins/logging.js
@@ -54,7 +54,7 @@ function formatter(id, level) {
 module.exports = function logging (server, conf) {
   var level = conf.logging && conf.logging.level && LOG_LEVELS.indexOf(conf.logging.level)
   if (level === undefined || level === -1 ) {
-    console.log('Warning, logging.level configured to an invalid value:', conf.logging.level)
+    console.log('Warning, logging.level configured to an invalid value:', level)
     console.log('Should be one of:', LOG_LEVELS.join(', '))
     level = DEFAULT_LEVEL
   }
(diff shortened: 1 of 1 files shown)


amsross:
These tests appear to be passing, but some other error is preventing the build from successfully completing.

Same issue in #441.

arj03:
@mmckegg good to merge?

plugins/logging.js
```diff
@@ -52,9 +52,9 @@ function formatter(id, level) {
 }
 
 module.exports = function logging (server, conf) {
-  var level = conf.logging && conf.logging.level && LOG_LEVELS.indexOf(conf.logging.level) || DEFAULT_LEVEL
-  if (level === -1) {
-    console.log('Warning, logging.level configured to an invalid value:', conf.logging.level)
+  var level = conf.logging && conf.logging.level && LOG_LEVELS.indexOf(conf.logging.level)
+  if (level === undefined || level === -1 ) {
+    console.log('Warning, logging.level configured to an invalid value:', level)
```
amsross: I think `level` should probably be `conf.logging.level` here now that I'm looking at it again.

`'Warning, logging.level configured to an invalid value: undefined'` doesn't seem to helpful.

In addition, not setting a log level will log this message, even though `conf.logging.level === undefined` is probably valid.

I can update this to reflect that thinking, but maybe I can get a sanity check real quick?

christianbundy:
Hey, sorry this wasn't merged -- was this resolved by #488?

christianbundy:
I think this is stale, and probably already resolved as #488. Please ping me if this should still be merged.

PR #444, opened 2017-10-14
replicate progress: don't include disconnected peers in incomplete count

mmckegg:
Fixes stuck downloading message progress in patchwork.

This (as far as I can possible see) is absolutely safe to merge and only affects Patchwork.

If no one complains within the next 24 hrs, I will merge it! I am already running this patch in patchwork (latest master) but it would be nice to get it into mainbot.

cc @dominictarr


dominictarr:
we should really update this to use the new style progress api. anyway, looks good, merging.

dominictarr:
merged into 10.4.10

PR #450, opened 2017-11-08
Make invite.accept work with onion

arj03:
Tested this together with https://github.com/ssbc/ssb-ref/pull/13 and it makes accepting invites over tor work again.


dominictarr:
@arj03 can you publish this as ... patch?

arj03:
@dominictarr what do you mean? Do a 1.4.11?

I'm still missing this patch for all my tor patches to be applied:

https://github.com/ssbc/secret-stack/pull/18

But that repo is in a wierd state, see https://github.com/ssbc/secret-stack/pull/16

Issue #451, opened 2017-11-20
indexes.forEach is not a function


fuzzyTew:
I just install scuttlebot and tried to accept an invite code.  The server immediately crashed and will not restart:
```
$ sbot server
flumelog-offset: blockSize and codec params moved into an object. https://github.com/flumedb/flumelog-offset
Log level: notice
.................................................. (indexes:0%)/usr/lib/node_modules/scuttlebot/node_modules/flumeview-query/index.js:25
    indexes.forEach(function (index) {
            ^

TypeError: indexes.forEach is not a function
    at /usr/lib/node_modules/scuttlebot/node_modules/flumeview-query/index.js:25:13
    at reduce (/usr/lib/node_modules/scuttlebot/node_modules/flumeview-level/index.js:88:25)
    at /usr/lib/node_modules/scuttlebot/node_modules/pull-write/index.js:35:23
    at /usr/lib/node_modules/scuttlebot/node_modules/pull-write/index.js:41:13
    at /usr/lib/node_modules/scuttlebot/node_modules/pull-cursor/cursor.js:32:13
    at getMeta (/usr/lib/node_modules/scuttlebot/node_modules/flumelog-offset/inject.js:43:14)
    at next (/usr/lib/node_modules/scuttlebot/node_modules/pull-cursor/cursor.js:27:11)
    at /usr/lib/node_modules/scuttlebot/node_modules/pull-cursor/cursor.js:20:11
    at Function.many.once (/usr/lib/node_modules/scuttlebot/node_modules/obv/index.js:39:7)
    at /usr/lib/node_modules/scuttlebot/node_modules/pull-cursor/cursor.js:13:13
````
Any ideas?

christianbundy:
Having the same problem on my first build. Unsure of whether I'm doing something wrong or whether this is a new regression.

fuzzyTew:
It seems the ssb-query module is instantiating the flumeview-query module with the value 1 for the indexes parameter, while flumeview-query is expecting this value to be an array.

The only way I can get sbot running is with the patchwork binary, and the modules are bundled up in it so I can’t inspect them to see what’s different.

dominictarr:
oops, I accedentially published a breaking change, fixed in flumeview-query@4.0.1 and ssb-links@2.0.1 and ssb-query@1.0.1, if you reinstall scuttlebot's node_modules you should get all this stuff.

kseistrup:
[...]

PR #452, opened 2017-11-22
Bring package-lock.json up-to-date

christianbundy:
It looks like this has fallen quite a bit behind, so it's hard to tell which packages should _actually_ be getting installed to get everything working. _Keeping_ this file up-to-date will depend on future PRs, but this is a one-time fix to bring us up-to-date.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,28 +1,330 @@
 {
   "name": "scuttlebot",
-  "version": "10.4.6",
+  "version": "10.4.10",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "async-single": {
+    "abstract-leveldown": {
+      "version": "0.12.4",
+      "resolved": "https://registry.npmjs.org/abstract-leveldown/-/abstract-leveldown-0.12.4.tgz",
+      "integrity": "sha1-KeGOYy5g5OIh1YECR4UqY9ey5BA=",
+      "requires": {
+        "xtend": "3.0.0"
+      },
+      "dependencies": {
+        "xtend": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/xtend/-/xtend-3.0.0.tgz",
+          "integrity": "sha1-XM50B7r2Qsunvs2laBEcST9ZZlo="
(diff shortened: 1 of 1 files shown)


mmckegg:
Or just delete it. 😆 

I am not a fan of `package-lock.json`.

Does anyone know of an article about the recommended practice around package-lock.json files? The only case I can think of where they make sense is deploying to a remote server. In app development, they are just a pain. Oh especially when some people are running old versions of npm that don't support them.

christianbundy:
https://docs.npmjs.com/files/package-lock.json

In lieu of an article, here's an example scenario:
1. We specify dependency foo@1.0.x.
2. Others depend on our package.
3. Foo accidentally pushes a breaking change to 1.0.1.
4. Every package that depends on ours is broken.

Lockfiles are a snapshot of the most recent working dependency tree, and they're _incredibly_ useful for issues like this: https://github.com/ssbc/scuttlebot/issues/451#issuecomment-346308695

We can (and probably should) specify which versions of Node and NPM work best with scuttlebot, and as long as you're using NPM to manage `package.json` you shouldn't have to make any workflow changes to keep the lockfile up-to-date.

mmckegg:
How often should a lock file be committed? Every time it changes?

What causes a lock file to update? Sorry I'm from the past 😆 and am still running npm v3. And I am scared of change. (and googling things)

Sorry about my tone above. I am genuinely interested in how lock files work.

christianbundy:
> How often should a lock file be committed? Every time it changes?

Yep, it's just like any other repo file. 

> What causes a lock file to update?

Running `npm install` after a manual change to `package.json`, `npm install --save`, or `npm update` should all do the trick.

> Sorry I'm from the past 😆 and am still running npm v3. And I am scared of change. (and googling things)

It was pretty hairy for a while, but _finally_ the lockfile is working as it should. Don't blame you for wanting to stay on a [more] stable version of NPM!

> Sorry about my tone above. I am genuinely interested in how lock files work.

No tone at all, amigo -- I'm usually the one chanting "delete it" any time a file starts causing trouble. 😉

PR #459, opened 2017-12-26
sbot server --logging.level=info output

donpdonp:
A stumbling block for me when trying sbot for the first time is knowing what (if anything) is going on with the server. These changes add config status and peer connect/disconnect/info messages at the 'info' logging level, and adds a README note to suggest starting with level info.

### DIFF
--- a/plugins/gossip/index.js
+++ b/plugins/gossip/index.js
@@ -282,18 +282,6 @@ module.exports = {
       if (!server.ready() && rpc.id !== server.id) return rpc.close()
 
       var peer = getPeer(rpc.id)
-      //don't track clients that connect, but arn't considered peers.
-      //maybe we should though?
-      if(!peer) {
-        if(rpc.id !== server.id) {
-          server.emit('log:info', ['SBOT', rpc.id, 'Connected'])
-          rpc.on('closed', function () {
-            server.emit('log:info', ['SBOT', rpc.id, 'Disconnected'])
-          })
-        }
-        return
-      }
-
       status[rpc.id] = simplify(peer)
 
       server.emit('log:info', ['SBOT', stringify(peer), 'Connected'])
--- a/plugins/gossip/index.js
+++ b/plugins/gossip/index.js
@@ -306,7 +306,7 @@ module.exports = {
 
       status[rpc.id] = simplify(peer)
 
-      server.emit('log:info', ['SBOT', stringify(peer), 'CONNECTED'])
+      server.emit('log:info', ['SBOT', stringify(peer), 'PEER JOINED'])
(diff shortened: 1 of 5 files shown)


plugins/gossip/index.js
```diff
@@ -191,6 +196,10 @@ module.exports = {
 
       }, 'string|object'),
 
+      handshakeTimeout: function(peer) {
```
dominictarr: I don't think we should add another method just to have a logging event, (new methods should only be added very reluctantly!)

How do you feel about adding this as either an 'reason' option to disconnect, (which also causes a message to be logged), or to just log from gossip/schedule?

donpdonp: I felt the handshake timeout was a 'real' event, ie. something that changed the state-machine of the client, so it seemed okay to make a new method. For example, what if the server did not want to disconnect on handshake timeout, but resend a hello message or otherwise restart the handshake without loosing the client connection.

A reason field on disconnect also seems fine (though I dont immediately understand the , 'string|object' syntax. grin. ), which ever you feel is best.

dominictarr:
okay, I'm gonna merge this and just change [this bit](https://github.com/ssbc/scuttlebot/pull/459/files#r159034571)

thank you for your pull request, it was especially easy to come to the decision to merge because you clearly expressed "A stumbling block for me when trying sbot for the first time is knowing what (if anything) is going on with the server."

To any one else reading this: it helps enormously to know the goal of the PR please don't just submit code!

plugins/gossip/index.js
dominictarr: oh yeah, some useful context: I think schedule shouldn't really be calling disconnect, that may be a hack because something else doesn't error properly. Also, if the timeout fails, it would be because of the tcp connection, and the thing that fails is either the tcp connection or the packets took too long in secret-handshake (authenticated encryption for that socket) it's hard to distinguish between a really slow tcp connection and a broken one, but in the p2p system it needs to be robust to breakages anyway, so just connect to another peer instead.

dominictarr:
merged into 10.5.0

donpdonp:
Great, thanks for merging the PR. I think scuttlebot is a very interesting and accessable platform for learning about distributed log systems. I hope the extra output here helps people get an idea of whats happening in the server. Peer joins/parts are just one area it'd be nice to have realtime info on. Messages in/out rate would be another.

dominictarr:
check `sbot status` - it has some information about that, but i am afraid I wrote it to help diagnose problems and it didn't get much attention after that. also, a rough ui here: https://github.com/dominictarr/ssb-status

PR #462, opened 2018-01-11
add start script

evbogue:
This way I can start scuttlebot instead of seeing an error when I habitually type `npm start` in the scuttlebot folder.


dominictarr:
merged, doesn't need an npm release, since you only npm start things you git cloned

PR #473, opened 2018-03-12
Adds way to create new sbots.

pietgeursen:
I need the ability to create multiple sbots, especially for testing.
This doesn't break the existing api. Just adds a `createSbot` function to the returned sbot object.

### DIFF
--- a/index.js
+++ b/index.js
@@ -155,11 +155,16 @@ function usage (cmd) {
   return mdm.usage(apidocs._, cmd)
 }
 
-module.exports = SecretStack({
-  //this is just the default app key.
-  //it can be overridden by passing a appKey as option
-  //when creating a Sbot instance.
-  appKey: require('./lib/ssb-cap')
-})
-.use(SSB)
+function createSbot() {
+  return SecretStack({
+    //this is just the default app key.
+    //it can be overridden by passing a appKey as option
+    //when creating a Sbot instance.
+    appKey: require('./lib/ssb-cap')
+  })
+    .use(SSB)
+}
+const stack = createSbot()
+stack.createSbot = createSbot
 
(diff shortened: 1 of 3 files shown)


evbogue:
@pietgeursen Someone was asking about running two sbots at the same time with different network keys, does this make it possible to do that?

Also can you give a quick example of how to use this to start an sbot?

dominictarr:
I think you can already do this, here is how it happens in the sbot tests

https://github.com/ssbc/scuttlebot/blob/master/test/block3.js#L20

pietgeursen:
@evbogue 

> Someone was asking about running two sbots at the same time with different network keys, does this make it possible to do that?

Yeah I'm pretty sure it does. It creates a whole new secret stack object which is what I think you need to be able to do that. 

> Also can you give a quick example of how to use this to start an sbot?

Yep, I added some tests https://github.com/ssbc/scuttlebot/pull/473/files#diff-73462efcdc10279b7ac47098953134a9

See also https://github.com/ssbc/scuttle-testbot which is how I found this problem.

@dominictarr 

> I think you can already do this, here is how it happens in the sbot tests

See the tests I just added in this PR. The painful part of the current api is that when you have multiple files requiring in scuttlebot, you don't know who has already set up which plugins.

dominictarr:
oh, I see what you mean now. you mean to create multilpe scuttlbot instances with _different plugins_.
Gonna merge, but we don't need a test that the current api throws

dominictarr:
merged into 11.1.0

pietgeursen:
Rad! Thanks for moving so fast on this. Any reason why you didn't publish?

dominictarr:
published as 11.1.1

PR #475, opened 2018-03-14
Update deps

arj03:
I was testing charwise (http://between-two-worlds.dk:8807/channel/benchmark-ci) and noticed that ssb-query 1.0.0 includes an ancient streamview-links. There is a 1.0.2 version but also a 2.0.0. The package-lock.json file seems to pin things to 1.0.0 and it has been resurrected in PR #473.

Related #468.


cryptix:
+1 on the ssb-query update!

I don't know how to feel about the package-lock.json dance.. It would be good to have a clear strategy when and how to update it. A package.json dependency change seems like the point in time when to sync it up and commit it.

optikfluffel:
Maybe using something like [greenkeeper.io](https://greenkeeper.io) could help to always stay up to date.

dominictarr:
I've found package-lock to be generally a nusance. It makes the diffs really big... I'd lean towards just removing the package-lock.json

staltz:
Between big diffs and deterministic installs, I would choose deterministic installs. The way of getting DI without committing package-lock is by fixing all versions in our dependencies, but that still isn't enough since it doesn't apply to transitive dependencies

evbogue:
I'm for commiting `package-lock.json` files. It makes the diffs a mess, but maybe we could get around that by making `package-lock.json` updates their own commit? 

The biggest reasons for this is it cuts down on the number of unexpected problems that newcomers to ssb face, as we'd be the ones who are making sure new patches are working.

dominictarr:
update: package-lock is only really useful for applications, not modules. so to actually use package-lock, in recent versions of npm, you'd have to clone scuttlebot, and then use `npm install --package-lock-only` (because we have ^ version ranges). If you do `npm install scuttlebot -g` that doesn't use package-lock either, I figure that is what most people do. 

Maybe the answer is to bundle scuttlebot into another thing, and use package-lock in that, then bundle all deps, so that npm installs just that. (although that would make one giant dep, with all the unchanging deps reinstalled each time)

Okay, so my official policy for this: is modules do not need package-lock, scuttlebutt is borderline, but package-lock in patch{work,bay,less} is a good idea

PR #478, opened 2018-03-21
Include server name in connection error

tantalum:
Include the server name scuttlebot is trying to connect to when displaying the error indicating that the connection failed.


dominictarr:
merged, thanks! sorry for the wait

PR #482, opened 2018-04-05
Added `id` flag for clarity.

austinfrey:
When using `createHistoryStream` programmatically or `hist` from the CLI, it does not appear that passing the feedId as a string works. Passing in an object `{id: <feedId>}` or the `--id` flag using the CLI does though. Updated the readme for clarity. 

Currently running Ubuntu 16.+

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ sbot feed
 sbot log
 
 # stream all messages by one feed, ordered by sequence number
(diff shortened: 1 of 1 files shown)


dominictarr:
thanks! merged!

dominictarr:
merged into 11.1.2

Issue #484, opened 2018-04-09
TypeError: Cannot read property 'sync' of null


cryptix:
I fear I found a reason for bad pub lifetime after trying andyleaps go port.

I tried it with two other sbots, and they both crash pretty soon.. possibly because it gets a malformed message during sync?

```
Apr 09 15:05:49 apu42 bin.js[2568]: /home/sbot/scuttlebot/lib/detect-sync.js:20
Apr 09 15:05:49 apu42 bin.js[2568]:     if (msg.sync) {
Apr 09 15:05:49 apu42 bin.js[2568]:             ^
Apr 09 15:05:49 apu42 bin.js[2568]: TypeError: Cannot read property 'sync' of null
Apr 09 15:05:49 apu42 bin.js[2568]:     at /home/sbot/scuttlebot/lib/detect-sync.js:20:13
Apr 09 15:05:49 apu42 bin.js[2568]:     at /home/sbot/scuttlebot/node_modules/pull-stream/throughs/through.js:17:24
Apr 09 15:05:49 apu42 bin.js[2568]:     at source (/home/sbot/scuttlebot/node_modules/muxrpc/pull-weird.js:46:30)
Apr 09 15:05:49 apu42 bin.js[2568]:     at /home/sbot/scuttlebot/node_modules/pull-stream/throughs/through.js:16:14
Apr 09 15:05:49 apu42 bin.js[2568]:     at next (/home/sbot/scuttlebot/node_modules/pull-stream/throughs/async-map.js:23:9)
Apr 09 15:05:49 apu42 bin.js[2568]:     at next (/home/sbot/scuttlebot/node_modules/pull-stream/sinks/drain.js:16:11)
Apr 09 15:05:49 apu42 bin.js[2568]:     at /home/sbot/scuttlebot/node_modules/pull-stream/sinks/drain.js:29:15
Apr 09 15:05:49 apu42 bin.js[2568]:     at Immediate._onImmediate (/home/sbot/scuttlebot/node_modules/pull-stream/throughs/async-map.js:35:20)
Apr 09 15:05:49 apu42 bin.js[2568]:     at runCallback (timers.js:794:20)
Apr 09 15:05:49 apu42 bin.js[2568]:     at tryOnImmediate (timers.js:752:5)
```

I patched lib/detect-sync.js:20 to just ignore null messages for now..

PR #485, opened 2018-04-09
hack: don't use null messages

cryptix:
hack as in i'm not 100% sure why but at least my sbot doesn't crash

see ssbc/scuttlebot#484

### DIFF
--- a/lib/detect-sync.js
+++ b/lib/detect-sync.js
@@ -17,11 +17,17 @@ module.exports = function detectSync (peerId, upto, toSend, peerHas, onSync) {
   }, 500)
 
   return function (msg) {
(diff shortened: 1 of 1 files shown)


dominictarr:
merged into 11.1.2

cryptix:
thank you!

do you mean _legacy_ as in not ebt?

> when did this start happening? 

i just recompiled the latest master of andyleap/go-ssb and it took down my sbots at each sync. I don't think many people use that code though but still.. poor sbots having to reboot are not good.

dominictarr:
yeah, there shouldn't be crashes because someone sends invalid data

PR #488, opened 2018-04-14
Fix logging for level 'error'

jfr3000:
sbot would ignore the logging level when it was set to "error" because its index was tested for falsiness. Now we separate testing for the existence of a config log level and getting its `LOG_LEVELS` index.

### DIFF
--- a/plugins/logging.js
+++ b/plugins/logging.js
@@ -52,8 +52,11 @@ function formatter(id, level) {
 }
 
 module.exports = function logging (server, conf) {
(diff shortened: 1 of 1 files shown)


jfr3000:
Whoops, I didn't see https://github.com/ssbc/scuttlebot/pull/440 which is still open with comments. This one does the correct thing though.

christianbundy:
Hey, sorry this took so long for someone to look at. The bug makes sense to me: when `LOG_LEVELS.indexOf(conf.logging.level) === 0` then it trips the `||` condition and goes back to the default. Thanks for fixing this, I'm going to merge your change now! :tada:

PR #494, opened 2018-04-29
work with https://github.com/ssbc/secret-stack/pull/21

dominictarr:
changes needed to be compatible with https://github.com/ssbc/secret-stack/pull/21
these changes only affect the invite plugin.


plugins/invite.js
```diff
@@ -183,6 +189,8 @@ module.exports = {
             modern = true
         }
 
+        console.log(invite)
+        console.log(ref.parseInvite(invite))
```
staltz: Could we use the logging plugin instead?

dominictarr:
sorry, fixed in 15acc86df0549ae9545bb23e2f829194eba30d12

PR #499, opened 2018-05-08
Fix Travis CI

### LINKED ISSUE #447: Travis tests/builds always fail
Travis builds are consistently failing. This has been the case for as far back as I went (which was not all that far) in the [build history](https://travis-ci.org/ssbc/scuttlebot/builds).

```
"test-plugin" has been installed. Restart Scuttlebot server to enable the plugin.

ok 1 (unnamed assert)
# installed and enabled plugin is loaded
ok 2 (unnamed assert)
ok 3 (unnamed assert)
ok 4 should be equal
# installed and disabled plugin is not loaded
ok 5 should be equal
# uninstall plugin
uninstalling plugin...

events.js:160
      throw er; // Unhandled 'error' event
      ^
OpenError: IO error: lock /tmp/test-plugins1/db/LOCK: already held by process
    at /home/travis/build/ssbc/scuttlebot/node_modules/level-packager/node_modules/levelup/lib/levelup.js:117:34
    at /home/travis/build/ssbc/scuttlebot/node_modules/leveldown/node_modules/abstract-leveldown/abstract-leveldown.js:39:16
npm ERR! Test failed.  See above for more details.

The command "npm test" exited with 1.

Done. Your build exited with 1.
```

christianbundy:
I've noticed that the Travis CI builds seem to be broken by the fact that [Travis doesn't support IPv6](https://docs.travis-ci.com/user/reference/precise/#Networking). This PR does three things:

- Disable IPv6 under Travis CI, emitting a simple warning instead
- Replace Node.js 6.9.5 test with Node.js 6.x for maintenance LTS testing
- Add Node.js 8 test with Node.js 8.x for active LTS testing

Please let me know if this could be done better, I'm open to any input. Thanks!

Resolves #447 

---

Note: it's still [possible](https://travis-ci.org/ssbc/scuttlebot/jobs/376165375) to fail, probably due to a race condition, but at least those jobs can be re-triggered with a reasonable probability of passing the next time. The goal of this PR is to make testing better, not perfect. :upside_down_face:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
 sudo: false
 node_js:
-  - '6.9.5'
+  - 8
+  - 6
--- a/test/invite.js
+++ b/test/invite.js
@@ -135,8 +135,8 @@ tape('test invite.accept doesnt follow if already followed', function (t) {
 })
 
 if (process.env.TRAVIS === 'true') {
-  console.warn('IPv6 is unsupported under Travis CI, test skipped');
-  skipIPv6 = true;
+  console.warn('IPv6 is unsupported under Travis CI, test skipped')
+  var skipIPv6 = true
 }
 
 tape('test invite.accept api with ipv6', { skip: skipIPv6 }, function (t) {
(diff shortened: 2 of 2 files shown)


dominictarr:
oh, looks good.

dominictarr:
I'm gonna add node 10 too, once I've let this run once to see if it passes.

PR #508, opened 2018-06-04
bump ssb-query to 2.1.0

mixmix:
This is needed to bring scuttlebot up to speed with other applications in the the system in terms of vesion of index it's using for ssb-query. (I think this is only relevant for the `sbot server` command?)


dominictarr:
merged in 11.3.2

PR #509, opened 2018-06-09
bump flumelog-offset (performance)

mixmix:
Off the back of this PR on flumelog-offset https://github.com/flumedb/flumelog-offset/pull/11, thought this would be a good step 

I've not tested this in production yet

I wanted to commit the package-lock, but see you removed it @dominictarr . I've started a thread on ssb about this so it's referenceable %fLDEuIiCDBjljsO8DEF3K/G/DPnCs3BVQ8SC6c8FFYI=.sha256

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "deep-equal": "^1.0.1",
     "explain-error": "^1.0.3",
     "flumedb": "^0.4.2",
-    "flumelog-offset": "^3.2.6",
+    "flumelog-offset": "^3.3.1",
     "flumeview-reduce": "^1.3.13",
     "has-network": "0.0.1",
     "ip": "^0.3.3",


arj03:
Yeah good idea with the bump, I was thinking I'd wait until the flumedb and aligned-block-file patches are merged before doing a release of this. But come to think of it, smaller steps are always easier to debug. 

There is also an interesting conversation [here](https://github.com/flumedb/flumelog-offset/pull/12) about block-size that might lead to some performance improvements.

mixmix:
Haha, know that Dominic uses mechanical hard drive BECAUSE it's slow.
Forces you to write better code etc

dominictarr:
@mixmix there is no package-lock because instead we use a npm-shrinkwrap in https://www.npmjs.com/package/scuttlebot-release. then the dependencies are locked when you do an `npm install`. A package-lock is only used on local installs, so it doesn't really help for most people. 

hmm, also I don't think flumelog-offset should actually be a dependency of scuttlebot? it should be in secure-scuttlebutt.

dominictarr:
updated in secure-scuttlebutt@18.0.7, scuttlebot@11.3.3, and scuttlebot-release@11.3.3

mixmix:
Cool, i didn't know about scuttlebot-release.
Is scuttlebot-release what we should really be using in eg ticktack? Seems
like it should be.

I actually think the main way the majority of people use scuttlebot is
based on a local install, because the electron build process (i think) just
bundles a local install.
I would assert that a electron bundle scuttlebot is the most common use
case right now given patchwork's prevalence

PR #519, opened 2018-08-01
fixes installation for node v10! updated deps

wanderer:


cryptix:
I’m not sure you want muxrpc@7 - we encountered strange blocking behavior with it and rolled back to 6.4

wanderer:
@cryptix ok reverted muxprc

dominictarr:
oh, looks like some other PR fixed this. thanks though!

PR #520, opened 2018-08-02
relatedMessages: remove leftovers

### LINKED ISSUE #518: relatedMessages still handled
This method was removed in ssb@18 but There is still a wrapper for that call and some api docs that should be removed.

See: http://viewer.scuttlebot.io/%25aXka56%2BSBANy6iKejn2cqZ%2Bgbyz5mc3HatFNGg5oBR8%3D.sha256

cryptix:
this call was removed in secure-scuttlebutt v18.

fixes #518

### DIFF
--- a/api.md
+++ b/api.md
@@ -199,44 +199,6 @@ The objects in this stream will be of the form:
  - `values` (boolean, default: `true`): whether the `data` event should contain values. If set to `true` and `keys` set to `false` then `data` events will simply be values, rather than objects with a `value` property.
 
 
-
-## relatedMessages: async
-
-Retrieve the tree of messages related to the given id.
-
-```bash
-relatedMessages --id {msgid} [--rel value] [--count] [--parent]
-```
-
-```js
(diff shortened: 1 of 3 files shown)


cryptix:
I'm not so sure if/how many blank lines are required in the `api.md` but otherwise this looks self explanatory to me and the tests still pass.

arj03:
Looks good, there is also cleaning up http://scuttlebot.io/apis/scuttlebot/ssb.html which is probably just as important.

cryptix:
Right! Do you know which repo that is? It’s not ssbc.github.io and not the handbook...

dominictarr:
@cryptix thanks! merged into 11.4.2

arj03:
@cryptix that should be https://github.com/ssbc/scuttlebot.io I think.

cryptix:
Yup. right. Thanks!

I just need to figure out what to do with https://scuttlebot.io/docs/advanced/post-threads.html

I guess links2 is the best api for that? I'll make a new thread for that question.

PR #521, opened 2018-08-02
invite: make sure we pass a string to ip.isPrivate

### LINKED ISSUE #517: Cannot read property 'match' of undefined
match is undefined on path "/scuttlebot/node_modules/ip/lib/ip.js"
on line number 262 addr.match

cryptix:
fixes #517


cryptix:
Would prefer to have a test case for their root issue but I don't know why it's occurring but this should help a little...

plugins/invite.js
```diff
@@ -76,6 +76,9 @@ module.exports = {
 
         var addr = server.getAddress()
         var host = ref.parseAddress(addr).host
+        if(typeof host !== 'string') {
+          return cb(new Error('Could not parse host portion from server address:', addr))
```
staltz: Shouldn't it be `new Error(a + b)` instead of `new Error(a, b)`? I guess, not sure, because the signature of this is `new Error([message[, fileName[, lineNumber]]])`

cryptix: thank you! I was assuming string concat like in `console.log`.

christianbundy:
@cryptix Sorry I haven't seen this until now, is this still good to merge? I see a merge conflict but otherwise this LGTM.

PR #525, opened 2018-08-19
Prep work to pass config down to broadcast for ipv6

jedahan:
Mostly just passing config along, broadcast does the hard work :)

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -22,7 +22,12 @@ var i = argv.indexOf('--')
 var conf = argv.slice(i+1)
 argv = ~i ? argv.slice(0, i) : argv
 
-var config = require('ssb-config/inject')(process.env.ssb_appname, minimist(conf))
+var config = require('ssb-config/inject')(
+  process.env.ssb_appname,
+  Object.assign(
+    minimist(conf),
+    { host: process.env.ssb_host }
+  ))
 
 var keys = ssbKeys.loadOrCreateSync(path.join(config.path, 'secret'))
 if(keys.curve === 'k256')
--- a/plugins/local.js
+++ b/plugins/local.js
@@ -34,7 +34,7 @@ module.exports = {
         }
       }
 
-    var local = broadcast(config.port)
+    var local = broadcast(config.port, config.loopback, config.family)
     var addrs = {}
     var lastSeen = {}
 
@@ -54,6 +54,7 @@ module.exports = {
       var data = buf.toString()
       var peer = ref.parseAddress(data)
       if (peer && peer.key !== sbot.id) {
+        if (config.family === 'IPv6') peer.host = buf.address
         addrs[peer.key] = peer
         lastSeen[peer.key] = Date.now()
         sbot.gossip.add(data, 'local')
--- a/index.js
+++ b/index.js
@@ -166,6 +166,7 @@ function createSbot() {
     //this is just the default app key.
     //it can be overridden by passing a appKey as option
     //when creating a Sbot instance.
+    host: process.env.ssb_host,
     appKey: require('./lib/ssb-cap')
   })
     .use(SSB)


christianbundy:
Hey @jedahan, sorry this hasn't been reviewed yet.

Could you give some further context on what this change is and why it's beneficial? It looks like it now takes a new environment variable, `ssb_host`, plus an IPv6 change, but I don't think I understand why this is necessary (or where it's documented). Thanks a lot, looking forward to getting this resolved!

christianbundy:
Is this something that's important to get merged? It looks like there are some merge conflicts and I haven't seen an update in a while. Looking forward to updates!

bin.js
```diff
@@ -22,7 +22,12 @@ var i = argv.indexOf('--')
 var conf = argv.slice(i+1)
 argv = ~i ? argv.slice(0, i) : argv
 
-var config = require('ssb-config/inject')(process.env.ssb_appname, minimist(conf))
+var config = require('ssb-config/inject')(
+  process.env.ssb_appname,
+  Object.assign(
+    minimist(conf),
+    { host: process.env.ssb_host }
```
dominictarr: don't put a hack like this here... this should already work in ssb-config. are you sure it's needed?

plugins/local.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
         }
       }
 
-    var local = broadcast(config.port)
+    var local = broadcast(config.port, config.loopback, config.family)
```
dominictarr: what does this do?

dominictarr: oh https://github.com/dominictarr/broadcast-stream/blob/master/index.js#L6 loopback makes the broadcast detectable on the same computer (needed for testing)

Was there a problem with this on ipv6? to OP mentions ipv6 but doesn't state it clearly.

bin.js
jedahan: It may no longer be needed, at the time I think this was a side-effect of how ssb_host was being read / overwritten.

plugins/local.js
jedahan: So it will be a little bit of time before I can get a testing setup up and running again. The motivation for this PR was to allow for nodes who are on the same physical network, but only have link-local (self-assigned) IPV6 addresses, to discover each other using another plugin I had written, using ipv6 multicast (similar to ipv4 broadcast).

I need to look into it again, but I think I had to enable loopback to not filter messages from other hosts, not just self-broadcasts.

I opted to pass in the family so the code I was submitting didn't disrupt the normal code path, though it may be better for both to behave the same was, as seen below.

plugins/local.js
```diff
@@ -54,6 +54,7 @@ module.exports = {
       var data = buf.toString()
       var peer = ref.parseAddress(data)
       if (peer && peer.key !== sbot.id) {
+        if (config.family === 'IPv6') peer.host = buf.address
```
jedahan: This is the crux of things - if a node is saying 'I am a node' it feels weird to me to use the contents of their message as the IP - why not use the address the packet came from? (From what I understand, its so my node can forward publicly routeable advertisements, which makes sense). With ipv6 link local, addresses are unique to the interface, so you cannot share another nodes address and it be able to connect.

plugins/local.js
dominictarr: you should have said something like this in the original post! I think the main reason this PR has languished is because it wasn't clear what you where trying to accomplish, so it wasn't clear how important this was, and so we soon got distracted by other things.

The biggest problem with this code is that it's really hard to test. To really cover all the sorts of network configurations we'd need a much more complicated testing set up than we have currently, like spinning up virtual networks running in VMs. This stuff is currently mainly QA'd by people with unusual network set ups complaining that it's not working.

can you give more context about how/why you are using ipv6? is ipv6 mission critical for you or nice to have?

jedahan: The goal is to be able to allow nodes to join a network without relying on a dhcp server. The 'easiest' way I've found for this is with self-assigned ipv6 addresses, since they won't really collide. I am open to figuring out other ways, but my knowledge is limited.

This is for mainly for two projects:

1. https://baculus.co - an experiment in disaster response infrastructure
2. An art project with a friend related to 'the distributed web of care' where we have participants connecting ad-hoc, though I think I can work around this by only having one node in the room act as a dhcp server.

dominictarr: right, because dhcd is completely different on ipv6 because you can just choose your own ip address. I've also managed to figure out that the reason it takes so long to connect to a wifi is because DHCP spec mandates multisecond delays

I'm definitely keen to support ssb on adhoc networks! another idea we've kicked around it a public open wifi, that lets you download and install an ssb client on the captive portal page

jedahan: Hmm I wonder if I should setup a video demo of baculus, it totally can do this. Right now it just serves up mvd since that requires zero installation from an end-user, but I should also expose Manyverse if it detects and android phone.

dominictarr: what is MVD? video demo is always good!

jedahan: https://github.com/evbogue/mvd - its a web client for scuttlebutt, that hosts identity keypairs in the browser. I'll make a video this week :)

dominictarr: oh right now I remember

plugins/local.js
cryptix: I also don’t get it..

https://github.com/ssbc/ssb-server/issues/640#issuecomment-466277322

staltz:
@jedahan I recently worked on fixing multiserver to support link-local IPv6 addresses: https://github.com/ssbc/multiserver/pull/52

Would that be sufficient to support the use cases you want, or does this PR still stand as necessary?

jedahan:
@staltz I think that is enough. I will test it now. Is there a ssb dev setup guide somewhere? It would be nice to see a checklist, something along the lines of:
* grab this container image
* setup a mini-network between nodes
* create a few ids
* put a test network in caps (or whatever its called now)
* mount a volume in some node_modules folder
* hack hack hack

christianbundy:
FWIW, I'm not familiar with a guide like that. However, you may be interested in: https://github.com/ssbc/ssb-integration-tests

PR #530, opened 2018-08-30
Respect hops for incoming and outgoing connections

arj03:
The idea with this pull request is that we shouldn't allow connections with any peers outside our hops. As it is now, sbot can connect to anyone but will only sync feeds within our hops. 

The biggest reason for this change is that right now anyone can connect to any peer and start requesting all feeds and thus map the whole network and use this for whatever reason they might have.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -60,6 +60,26 @@ if (argv[0] == 'server') {
   config.keys = keys
   var server = createSbot(config)
 
+  server.on('rpc:connect', function (rpc, isClient) {
+    if (rpc.id == '@' + keys.public) return
+
+    pull(
+      server.friends.hopStream(),
+      pull.drain(s => {
+        try {
+          var hops = s[rpc.id]
+          if (hops == undefined || hops > config.friends.hops) {
+            server.emit('log:info', ['SBOT', 'closing connection outside hops: ', hops])
+            rpc.close(true, function() { })
+          }
+        } catch (e) {
+          server.emit('log:info', ['SBOT', 'closing connection outside hops'])
+          rpc.close(true, function() { })
+        }
+      })
+    )
+  })
+
   // write RPC manifest to ~/.ssb/manifest.json
   fs.writeFileSync(manifestFile, JSON.stringify(server.getManifest(), null, 2))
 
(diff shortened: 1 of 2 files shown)


bin.js
```diff
@@ -60,6 +60,26 @@ if (argv[0] == 'server') {
   config.keys = keys
   var server = createSbot(config)
 
+  server.on('rpc:connect', function (rpc, isClient) {
+    if (rpc.id == '@' + keys.public) return
+
+    pull(
+      server.friends.hopStream(),
```
cryptix: do we have to worry about dangling streams of these?

will they emit values after the connection has been handled?

bin.js
```diff
@@ -60,6 +60,26 @@ if (argv[0] == 'server') {
...
+      pull.drain(s => {
+        try {
+          var hops = s[rpc.id]
+          if (hops == undefined || hops > config.friends.hops) {
+            server.emit('log:info', ['SBOT', 'closing connection outside hops: ', hops])
+            rpc.close(true, function() { })
+          }
+        } catch (e) {
```
christianbundy: Wow, this is really cool. I appreciate you taking this on. I have a couple of questions, just for my better understanding:

- Which sorts of errors can we expect to happen within this `try {}` block?
- Would `hops === undefined` have the same effect here? Or `hops == null`?
- Can `rpc.close()` be called with only one argument? I'm not sure if [this](https://github.com/ssbc/scuttlebot/blob/f6e2d256794d075ec47526344d41a008b5a814bd/index.js#L61) is the right code, but it looks like the argument could be optional.

staltz:
This is related to a PR I might do soon (but been having difficulties testing in practice): don't do RPC connections to peers you have *blocked*. It's related to hops, but I think blocking being opt-in wouldn't intersect with the invite use case.

arj03:
@staltz yes. Once I get to get bottom of this, I'll be sure to share pointers :)

arj03:
@staltz I think dominic actually already [fixed](https://github.com/dominictarr/epidemic-broadcast-trees/commits/master) that case in epidemic-broadcast-tress 6.3.4.

PR #531, opened 2018-08-31
Fix typo

arj03:
This should be an easy fix. Simple typo. Then again, I'm not really sure how that can be called twice looking at the code.


staltz:
Yes, an obvious merge

PR #532, opened 2018-08-31
Remove old and unused block api doc

arj03:
This is really old and not used anymore.


christianbundy:
I *think* this might be ripe for removal too: https://github.com/ssbc/scuttlebot/blob/f11eacb2457ed757f2b267720f56e33fd206c42f/lib/apidocs.js#L6

PR #535, opened 2018-09-03
WIP Replace gossip plugin with sbot-connection-manager

pietgeursen:


plugins/gossip.js
```diff
@@ -0,0 +1,281 @@
...
+var feedIdRegex = new RegExp(ref.feedIdRegex)
+
+function toMultiserverAddress (address) {
+  if (isObject(address)) {
+    if (ref.isFeed(address.key)) {
+      address.key = address.key.match(feedIdRegex)[1]
+    }
+    var protocol = 'net'
+    if (address.host.endsWith('.onion')) { protocol = 'onion' }
+    return [protocol, address.host, address.port].join(':') + '~' + ['shs', toBase64(address.key)].join(':')
```
staltz: This piece of code (I think it's old from the previous code) assumes the existence of port and host in the address object. We should eventually not assume that from every multiserver address.

dominictarr:
@pietgeursen please don't make PRs without a description. This is a valuable opportunity to communicate context about what/how/why you've done what you've done. Just because we've talked about it in other channels does not mean that you can leave this out because there are other people who might see this, or in the future we might want to look back and understand the context this was created under.

pietgeursen:
@dominictarr Sure. 
I'm going to close this for now then. I'm picking this work up again on Monday so I'll re-open it with a description.

Written text is hard to get tone. But for me the tone of this seems grumpy. If you're annoyed at me about something (boring scuttlebutt posts?) let's have a direct conversation. Happy to jump on a call if that's useful.

If it's not grumpy, all good.

dominictarr:
I'm trying not to be grumpy, but one thing that makes me grumpy is PRs without describing context.
I mean, if I get a PR without a description, I'll ask what it's about before looking at the code.
I even include context when making a PR to my own modules (eg. https://github.com/ssbc/ssb-validate/pull/10 https://github.com/dominictarr/pull-reader/pull/7 )

PR #536, opened 2018-09-04
WIP/Tracking: connections overhaul / Multiserver config scope

cryptix:
This depends on three PRs:

I tried to explain what this wants to enable in the ssb-config [readme](https://github.com/ssbc/ssb-config/pull/13/files#diff-04c6e90faac2675aa89e2176d2eec7d8).

* [x] https://github.com/ssbc/multiserver/pull/14
* [x] https://github.com/ssbc/ssb-config/pull/13 
* [x] https://github.com/ssbc/secret-stack/pull/23

@arj03 did most of the work here and I tested it with him a couple of weeks ago. We introduced multiserver on its own, which had a bad crashing side-effect but that should be fixed as of mutliserver@1.13.4.

Opening this for more visibility, tracking and (meta) discussion.

cc @pietgeursen


cryptix:
i rebased this on 11.4.4 and squashed the last of my commits which were juggling commits in the package.json.

pietgeursen:
@cryptix 
This looks cool! Just trying to get up to speed with it.
Little question: Did https://github.com/ssbc/multiserver/pull/14 change the api of multiserver? I was looking to see if the README changed but it hasn't.

arj03:
@pietgeursen plugins must now have scope defined.

dominictarr:
those crashes are fixed if this is using secret-stack@4.2.1
so I think this is mergeable now right?

I pushed a commit to update deps

arj03:
Excellent! Yeah looks like we are ready to go. Want to make this a 11.5 or 12.0 release?

arj03:
I checkout out this branch again and realized that we havn't added unix-socket by default, so I fixed that

dominictarr:
@arj03 does this break anything? I didn't think this is a breaking change... hmm I guess `getAddress` may return more than one concatenated address now, which could break things (that is why invites plugin had to change).

arj03:
Yeah, I'm mostly thinking og being careful as it is a big change internally in several modules. There is also connections, while not breaking is a major new feature. I can do the release and communication, just want to be clear on the version number as I got burned a bit with the multiserver release, thankfully stalz was there to clean things up ❤️

dominictarr:
@arj03 if in doubt, a breaking change is the safe option.

arj03:
Thanks! I'll see if I can get the release out tomorrow.

dominictarr:
awesome!

PR #544, opened 2018-09-20
Add `addUnboxer` to be used by plugins

christianbundy:
This gives plugins the ability to call [`addUnboxer()`](https://github.com/ssbc/secure-scuttlebutt/blob/beff8d063a20a1604ffe9110eb0d0f0911fc2e95/minimal.js#L187-L189) and add custom unboxer functions. This fits my use-case for blob-content posts, which use a [custom unboxer](https://github.com/ssbc/secure-scuttlebutt/blob/06937996465a317ac2c16ce234508900e1e69e56/minimal.js#L79-L106).

Relevant discussion on SSB: `%rciFuVmIAi6WxamBNcF+EYSnJtngAbHwxxmtudsz3v4=.sha256`

### DIFF
--- a/api.md
+++ b/api.md
@@ -339,3 +339,7 @@ an overview of how ssb is working.
 ## version: sync
 
 return the current version number of the running server
+
+## addUnboxer: sync
+
+add an unboxer to be used by the db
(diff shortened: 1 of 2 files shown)


dominictarr:
oh, this is actually a duplicate of https://github.com/ssbc/scuttlebot/pull/510 
hmm, we shouldn't actually add it to the `api.md` as that exposes it over muxrpc, and this can only be called within the same process, because you can't pass a function over muxrpc.

christianbundy:
Nice catch! I'll close this in favor of the original, thanks!

dominictarr:
this is now supported in 12.1.0

PR #550, opened 2018-10-02
gossip uses multiserver addresses locally

dominictarr:
I want gossip to use multiserver addresses locally, so that you can for example, gossip over websockets, or webrtc, or [tunnel](https://github.com/dominictarr/ssb-tunnel) to a peer. The code is still ugly, but I just fixed up the minimum amount to get this in.

Also I added an option to disable replication to `pub` advertisements. These were an ugly hack that we made back in the day, that I'd like to remove. (switching to user invites should achieve that) in particular, they are a problem because it's hard to tell if a pub is really still at that address - and since you announce someone else's address, which is insufficiently consentful.

### DIFF
--- a/plugins/gossip/index.js
+++ b/plugins/gossip/index.js
@@ -215,7 +215,7 @@ module.exports = {
       }, 'string|object'),
 
       disconnect: valid.async(function (addr, cb) {
-        var peer = this.get(addr)
+        var peer = gossip.get(addr)
 
         peer.state = 'disconnecting'
         peer.stateChange = Date.now()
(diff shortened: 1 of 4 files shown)


plugins/gossip/index.js
```diff
@@ -254,7 +282,7 @@ module.exports = {
         //between 10 seconds and 30 minutes, default 5 min
         timeout = Math.max(10e3, Math.min(timeout, 30*60e3))
         return ping({timeout: timeout})
-      },
+1      },
```
staltz: What's this `1` here?

PR #551, opened 2018-10-11
refactor networking plugins, don't use by default, and use ssb-ws instead of inlined ws plugin

dominictarr:
I started out just wanting to get the old ssb-ws plugin updated, and working with sbot again, but in 12 sbot added it's own ws server, so I needed to move that... and then I also thought that those other plugins should be moved away.

I'm seriously thinking that secret-stack shouldn't even load a networking plugin (or shs) and that should be loaded by sbot user...

I've also been thinking that now that it's so easy to add more protocols, and they are well-integrated, we need a good way to add protocols to ssb-client... it would be nice if ssb-client also had a `.use(plugin)` method - but it just didn't create a server? something like this is probably important to get ssb on the browser

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -39,6 +39,9 @@ if (argv[0] == 'server') {
   // import sbot and start the server
 
   var createSbot = require('./')
+    .use(require('./plugins/onion'))
+    .use(require('./plugins/unix-socket'))
+    .use(require('./plugins/no-auth'))
     .use(require('./plugins/plugins'))
     .use(require('./plugins/master'))
     .use(require('./plugins/gossip'))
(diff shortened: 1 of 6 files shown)


dominictarr:
@arj03 @regular

regular:
Nice!
@dominictarr I don't see your changes to [ssb-ws](https://github.com/ssbc/ssb-ws) Am I looking at the wrong repo or did you forget to push?

dominictarr:
sorry, pushed now

staltz:
I fully agree with this refactor PR as well as with secret-stack not loading transports & transforms by default.

arj03:
Yep agree. +1

PR #555, opened 2018-10-14
Remove unused dependencies, update others

christianbundy:
This removes all unused dependencies and brings the rest up-to-date. The only dependency left behind is muxrpc-validation, which throws errors about missing callbacks in our tests. This passes `npm test`, but has undergone no other testing.


cryptix:
oh.. i didn't mean to push b678e7faec6bb6badeb8405acadb5f244ac81512 onto this...

(I'd like to see if those trigger, though)

christianbundy:
Your commit LGTM, any chance you've run into bugs while testing this? It's been buttery smooth for me, but that's only a sample size of 1.

cryptix:
Has been working like a charm for me. Also used to run #557 - So as far as test coverage is an indicator, this should be fine, I think.

christianbundy:
Sweet, if you add an "approve" review I'll merge and we can let it bake in `master` for a while before the next release.

arj03:
Looks good @christianbundy!

PR #556, opened 2018-10-15
Fix config port and host

regular:
This bumps ssb-config to 2.3.5 to include the config.host, config.port bugfix
and adds tests for bin.js (sbot binary)
- test for config via command line (--host and --port, ipv4 + ipv6)
- test for config via config file

This ensures that the default `incoming` connection takes top-level config options into account.


regular:
I think travis fails because the ports I use for testing are not allowed on travis. It works fine locally.

regular:
You can merge #557 instead, it also contains these two commits.

PR #557, opened 2018-10-15
Test for bin.js

regular:
Add test for sbot getAddress 
sbot should provide a multiserver address containing ws: by default

Also bumps ssb-config dependency to 2.3.6, which will make the test pass.

NOTE: this PR also contains the two commits from my previous PR, so either just merge this one and close the other one, or merge the other one first, then this. You also need to publish ssb-config 2.3.6, there's a PR in ssb-config too.

### DIFF
--- a/test/bin.js
+++ b/test/bin.js
@@ -1,0 +1,235 @@
+var fs = require('fs')
+var test = require('tape')
+var spawn = require('child_process').spawn
+var exec = require('child_process').exec
+var crypto = require('crypto')
+var net = require('net')
+var mkdirp = require('mkdirp')
+var join = require('path').join
+var ma = require('multiserver-address')
+
+// travis currently does not support ipv6, becaue GCE does not.
+var has_ipv6 = process.env.TRAVIS === undefined
+
+function sbot(t, argv, opts) {
+  opts = opts || {}
+  // we spawn a shell with job control enabled
+  // that starts sbot on the background and then
+  // reads from stdin, which is a pipe to our node process
+  // When that pipe closes (because we do so when the tests ends, or
+  // it happens because our process dies), sbot will be killed
+  // by the shell automatically.
+  // Because the last command in the shell is 'wait %1', the shells' exit code
+  // will be sbot's exit code (which is >128 if it was killed)
+  var sh = spawn('bash', [
+    '-c', 
+    'set -o monitor; echo pwd: $(pwd); node ' + join(__dirname, '../bin.js') + ' ' +
+    argv.join(' ') +
+    ' & read dummy; echo killing sbot; kill %1; wait %1' 
+  ], Object.assign({
+    env: Object.assign({}, process.env, {ssb_appname: 'test'}),
+    stdio: ['pipe', 'inherit', 'inherit']
+  }, opts))
+
+  return function end() {
+    console.log('ending ...')
+  
+    sh.on('exit', function(code) {
+      if (code>128) {
+        t.comment('sbot was killed as expected')
+      } else {
+        t.fail('sbot exited with code ' + code)
+      }
+      t.end()
+    })
+    // closing shell's stdin will make it kill sbot
+    // if it is still running at this point.
+    // Either way, we'll get sbot's original exit code
+    // in the exit event above.
+    sh.stdin.end()
+  }
+}
+
+function try_often(times, opts, work, done) {
+  if (typeof opts == 'function') {
+    done = work
+    work = opts
+    opts = {}
+  }
+  const delay = 2000
+  setTimeout(function() { // delay first try
+    work(function(err, result) {
+      if (err) {
+        if (opts.ignore && err.message && !err.message.match(err.ignore)) {
+          console.error('Fatal error:', err)
+          return done(err)
+        }
+        if (times) return setTimeout(function () {
+          console.warn('retry run', times)
+          console.error('work(err):', err)
+          try_often(times-1, work, done)
+        }, delay)
+        return done(err)
+      }
+      return done(null, result)
+    })
+  }, delay)
+}
+
+test('run bin.js server with command line option --host and --port (IPv4)', function(t) {
+  var end = sbot(t, [
+    'server',
+    '--host=127.0.0.1',
+    '--port=9001',
+    '--ws.port=9002',
+    '--path=/tmp/sbot_binjstest_' + Date.now()
+  ])
+
+  try_often(10, function work(cb) {
+    var socket = net.connect(9001, '127.0.0.1')
+    socket.on('error', cb)
+    socket.on('connect', cb)
+  }, function done(err) {
+    t.error(err, 'Successfully connect eventually')
+    end()
+  })
+})
+
+if (has_ipv6)
+test('run bin.js server with command line option --host and --port (IPv6)', function(t) {
+  var end = sbot(t, [
+    'server',
+    '--host=::1',
+    '--port=9001',
+    '--ws.port=9002',
+    '--path=/tmp/sbot_binjstest_' + Date.now()
+  ])
+  try_often(10, function work(cb) {
+    var socket = net.connect(9001, '::1')
+    socket.on('error', cb)
+    socket.on('connect', cb)
+  }, function done(err) {
+    t.error(err, 'Successfully connect eventually')
+    end()
+  })
+})
+
+test('run bin.js server with local config file (port, host) (IPv4)', function(t) {
+  var dir = '/tmp/sbot_binjstest_' + Date.now()
+  mkdirp.sync(dir)
+  fs.writeFileSync(join(dir, '.testrc'), JSON.stringify({
+    host: '127.0.0.1',
+    port: 9001,
+    ws: {
+      port: 9002
+    }
+  }))
+  var end = sbot(t, [
+    'server',
+    '--path', dir
+  ], {
+    cwd: dir
+  })
+
+  try_often(3, {
+    ignore: /ECONNREFUSED/
+  }, function work(cb) {
+    var socket = net.connect(9001, '127.0.0.1')
+    socket.on('error', cb)
+    socket.on('connect', cb)
+  }, function done(err) {
+    t.error(err, 'Successfully connect eventually')
+    end()
+  })
+})
+
+if (has_ipv6)
+test('run bin.js server with local config file (port, host) (IPv6)', function(t) {
+  var dir = '/tmp/sbot_binjstest_' + Date.now()
+  mkdirp.sync(dir)
+  fs.writeFileSync(join(dir, '.testrc'), JSON.stringify({
+    host: '::',
+    port: 9001,
+    ws: {
+      port: 9002
+    }
+  }))
+  var end = sbot(t, [
+    'server',
+    '--path', dir
+  ], {
+    cwd: dir
+  })
+
+  try_often(10, function work(cb) {
+    var socket = net.connect(9001, '::1')
+    socket.on('error', cb)
+    socket.on('connect', cb)
+  }, function done(err) {
+    t.error(err, 'Successfully connect eventually')
+    end()
+  })
+})
+
+test('sbot should have websockets and http server by default', function(t) {
+  var path = '/tmp/sbot_binjstest_' + Date.now()
+  var caps = crypto.randomBytes(32).toString('base64')
+  var end = sbot(t, [
+    'server',
+    '--host=127.0.0.1',
+    '--port=9001',
+    '--ws.port=9002',
+    '--path', path,
+    '--caps.shs', caps
+  ])
+
+  try_often(10, function work(cb) {
+    exec([
+      join(__dirname, '../bin.js'),
+      'getAddress',
+      '--',
+      '--host=127.0.0.1',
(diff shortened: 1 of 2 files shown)


regular:
Oh, it's actually Google's fault that the tests involving IPv6 fail on travis!

From the [travis blog](https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future)

> With the move to GCE this will no longer work, until such time as as GCE adds IPv6 support. We do understand this may cause a disruption for some use cases and while we have considered numerous ways to try to provide IPv6 in the cloud, none of the current available options are suitable for a large production deployment.

regular:
🎉 Tests pass on Travis, finally! A little sad that we don't have coverage for IPv6 tests on travis though.

cryptix:
A little sad that we don't have coverage for IPv6 tests on travis though.

Agreed.. I understand that they can't provide connectivity but creating local addresses would be fine to cover lots of use cases, I believe.

cryptix:
Okay - AFAICT this just tests the `getAddress` result. I'll try to add another check about dialing the ws endpoint. Then we are good here, I think.

cryptix:
@regular If you don't object I'd squash some these commits together.

cryptix:
@regular I pushed a commit that establishes a connection to the ws part of the return of `getAddress`.

It connects but I'm getting _not on whitelist_ for everything I tried and I also tried to print the object. It's just empty....

I _thought_ that connecting with your own ID gives you the full API but it looks like it [now](https://github.com/ssbc/ssb-ws/blob/master/index.js#L60) just exposes API to people that you follow...

regular:
@cryptix Yes, some squashing is definitely required before merging.

Re ws connection testing: hm, I wouldn't mind if it didn't test if the websocket server actually works, because that's not something bin.js is responsible for, and this is a test for bin js. I would hope that there are tests for websockets somewhere else, or if they don't exist yet, I'd put them into ssb-ws and keep the scope of test/bin.js narrowed down on command line parsing and config file parsing (and whatever else happens in bin.js, blobs.add for example).

test/bin.js
```diff
@@ -0,0 +1,203 @@
...
+    }, function(err, stdout, sderr) {
+      t.error(err)
+      console.log('result of sbot getAddress', stdout)
+      const idx = stdout.indexOf('ws:')
+      t.ok(idx > 0, 'multiserver address should contain ws:')
+
+      require('ssb-client')(null, { // falsy value => load default from path
+        path: path,
+        caps: { shs: require('../lib/ssb-cap') }, // has to be set when setting any config
+        remote: stdout.substr(idx)
```
regular: this relies on the ws address to be the last one  within the multiserver address, that's not good.

test/bin.js
```diff
@@ -0,0 +1,203 @@
...
+})
+
+test('sbot should have websockets and http server by default', function(t) {
+  var path = '/tmp/' + Date.now()
+  var killed = false
+  var sbot = spawn(join(__dirname, '../bin.js'),  [
+    'server',
+    '--host=127.0.0.1',
+    '--port=8199',
+    '--ws.port=0', // use random port - get them from getAddress and dial that
```
regular: We need at least one test that confirms that config.ws.port is actually used. Letting it generate its own address partially defeats that purpose. (we know it checked for zero, we don't know if it uses non-zero port values)

cryptix:
> hm, I wouldn't mind if it didn't test if the websocket server actually works

In a sense I thought it is bin.js' responsability to load the plugin and supply it's service. Elaborate testing of it's own features should be done on ssb-ws itself but it felt incomplete to just check for an existing tcp port.

Right now ssb-ws has none though which is why I'm leaned towards a bit more integration testing. I also _think_ this shows the regression that `calling id == my id` worked before for full access. I'm not sure if this is related to ssb-friends 2 vs 3, specifically if `source == my == dest` was handled specially before. I agree though that all the different permutations should be checked over there.

test/bin.js
cryptix: Yes, agreed. I understand now. Using 0 to signal we want it to pick it's own won't tell us if it's broken and it just picks the default.

cryptix: In full transparency: I did this because sometimes the node process stuck around because the test script exited without killing it.

I'll try to use tape.onFinish to collect the PIDs and kill them collectively at the end.

cryptix:
I see now that you just did the net.connect in all the cases. I still think somehow that opening a TCP socket is not enough to see if that is the service we want (shs+muxprc or http+ws).

test/bin.js
regular: I see. We probably want to test both ws.getAddress and getAddress.

test/bin.js
regular: Ah, I see. Yes, that will do, as long as all the tests use different port numbers, so there's no ADDRINUSE when individual tests fail to kill sbot. It would be even better if we can have a finish handler that runs per test.

cryptix:
@regular I took the liberty to squash this a little already. I append the commit messages together. Hope this  is okay for you.

regular:
> I still think somehow that opening a TCP socket is not enough to see if that is the service we want (shs+muxprc or http+ws).

@cryptix fair point. I agree that it would be an improvement that the correct protocol is spoken on the connections.

Looks like you discovered another regression. Have you console.logged the keys in the client to compare with sbots output to confirm they are indeed identical?

cryptix:
I have it passing now. I also saw to passes on travis but [this](https://travis-ci.org/ssbc/scuttlebot/jobs/443118330#L488) error is quite persistent. I fear 1.2 seconds is to tight on those busy machines.

Since we can't get an event emitter to signal readyness through exec/spawn I'm leaning towards reading from stdout until it prints `my key ID:`

regular:
> Since we can't get an event emitter to signal readyness through exec/spawn I'm leaning towards reading from stdout until it prints my key ID

Yes, I thought about that too ... Then the test relies on an implementation detail of sbot though. How about a helper that retries 10 times with 1s pauses between tries? Something like:

```
function try_often(times, f, cb) {
  ff(function cb(err)) {
    if (err) {
       if (times) return try_often(times-1, f, cb)
       return cb(err)  
    }
    return cb(null)
  }
}
```

Maybe it should even take a regex for an error string that should be ignored and only repeats if that matches, so we can fail on any other error than whatever error we receive when sbot isn't ready yet.

cryptix:
AFAICT it's only _# run bin.js server with local config file (port, host) (v4)_ that fails on travis.

cryptix:
Interesting... Managed to reproduce it on a cleanly installed system. Not sure what the cause is yet.

regular:
@cryptix in the latest log, all the connect errors happen _after_ `# all done. stopping 3 bots, just to have no leftovers`

cryptix:
Oh.. sorry, don’t be confused by that number. I unconditionally kill them again at the end.

but I also don’t get why everything else passes. It shouldn’t kill that bot until the retry is reached or it just passes....

regular:
@cryptix it was surprisingly complex to make sbot die with the node process that runs the test. Now it seems to work though. (The trick is using a sub shell that reads from stdin and than kills the sbot, where stdin is piped from the node process that runs the test. if the test process dies because of an exception or whatever, the pipe is closed, the subshell's read doesn't block anymore and the sbot is killed. I was surprised to learn that there's no easier way to do this!)

I added a helper function that creates an sbot instance this way and returns a function to kill it and end the test. I could simplify a lot of things using it, like hard-coding the port numbers

I also implemented an option to try_often() to explicitly tell it what error(s) it can ignore (and then retry). other errors will make the test fail immediately.

i think we can merge this now, when travis gives a green light. My commit did not seem to trigger travis though. Might be the overall github brokenness today. :/

dominictarr:
I would be happy to make `close` a function callable over RPC if it made running tests like this easier.

hmm, I like your read from stdin trick @regular what if we added that to sbot also?

```
process.stdin.on('data', function () {}) //must have data listen else input is paused
process.stdin.on('end', function (err) {
  console.log('CLOSE')
  clearInterval(int) //close server
})

//this keeps the process up
var int = setInterval(function () {
  console.log(new Date())
}, 1000)
```
then we wouldn't need an ugly script... and having something like this is probably a good idea because otherwise undead separate process plugins are gonna be a major headache!

regular:
@dominictarr I am not sure if I get what you mean with the pseudo code, where would it live? It looks like some sort of heartbeat keep-alive watchdog thing. Is the idea, that the application that spawned sbot can tell it to shut down when it didn't hear back from the application for a while?

A more simple mechanism could be to make sbot/bin.js detect whether stdin is a pipe and then close down when stdin ends, so processes that spawn sbot can conveniently and robustly pull sbot with them (into the grave) when they die unexpectedly. That would put the mechanism from the shell script into sbot itself.

Yes, an explicit `close` would make life easier, not just for testing but also for using sbot in scripts. I think ssb-party is solving that via a plugin, but I don't see a reason why it shouldn't be built-in.

regular:
@cryptix the reason why the last test still fails is: the try_often worker must only do the getAddress call, the rest of the assertions and, most importantly, the end() call, must go into try_often `done`. I'll fix it.

dominictarr:
> A more simple mechanism could be to make sbot/bin.js detect whether stdin is a pipe and then close down when stdin ends, so processes that spawn sbot can conveniently and robustly pull sbot with them (into the grave) when they die unexpectedly

yes, that was what I was trying to illustrate ;) it was just a proof of concept test script... the setInterval would really be sbot's server(s)

regular:
@dominictarr Ah! Your code is a stand-in for sbot, got it! Unfortunately, as I just found out, it's not as robust as I had hoped for. The code in test/bin.js does work when the test throws an exception. However, when the test process is interrupted with `^C`, sbot survives. I don't know why that is. Anyway, it's good enough for travis.

regular:
@cryptix 
> All checks have passed

regular:
@cryptix It should be a patch release, I guess. clients that rely on websockets or non-default ports can than declare `scuttlebot>=13.0.1`

regular:
@cryptix from reading https://github.com/ssbc/scuttlebot-release build.sh, I think you first publish scuttlebot then clone https://github.com/ssbc/scuttlebot-release and run ./buildsh <version of scuttlebot you just released>

regular:
@cryptix The reason why I don't like the idea of not bumping the version number is: you would to have to publish as 13.0.0-1. The semantics of '-1' are less well defined than 13.0.1  (the `-n` postfix, to my knowledge just means: I messed up the previous publish, let me fix it, where as `.1` means "I fixed something that should have worked before but did not)

cryptix:
agreed. 13.0.0 _might_ work now while .1 signals _okay this is fixed_.

regular:
Yes, and scuttlebot-release will _not_ pick up updates automatically, because it is shrinkwrapped, so it needs to be re-shrinkwrapped and re-published and it's version number needs to match the scuttlebot version number. This dance is worth it however, because it's the only way to guarantee that this particular combination of dependencies' versions are testes together. Super important for production use.

dominictarr:
> However, when the test process is interrupted with ^C, sbot survives. I don't know why that is. Anyway, it's good enough for travis.

why does unix even work like this? one time I took a job interview at google (I didn't really want to work there, just to get a glimpse inside the beast, so I took the interview) I was able to answer lots of questions about unix syscalls just from knowing the node.js apis, but one question got me: what happens to a child process when the parent process dies? I didn't know, but I guessed: If I had designed it, i'd make the child die too. to make a daemon you'd have to have a special system process adopt your child. seems simple and obvious, but no, it's the opposite of that.

cryptix:
> i'd make the child die too. to make a daemon you'd have to have a special system process adopt your child.

yup, that's the behavior I know and depend on in couple of setups. The latter I know as _disown_ing. Without it, stuff you start in your shell goes away when you kill or exit that.


In that one commit, when I tried to `^C` the test runner, it wouldn't actually exit until I manually killed the sbot inside, then everything folded away... Maybe I should have taken more notice of that but I was in a tight tinker>try>goto1 loop back then, just trying to get make it pass.

regular:
@dominictarr @cryptix agreed, it violates the Law of Least Astonishment. @cryptix I had to look it up, but as I just learned: the behaviour we know and expect is implemented in the  shell, not on the system level. From `man bash`

```
    The shell exits by default upon receipt of a SIGHUP.   Before  exiting,
       an  interactive  shell  resends  the  SIGHUP  to  all  jobs, running or
       stopped.  Stopped jobs are sent SIGCONT to ensure that they receive the
       SIGHUP.   To  prevent the shell from sending the signal to a particular
       job, it should be removed from the jobs table with the  disown  builtin
       (see  SHELL  BUILTIN  COMMANDS  below)  or marked to not receive SIGHUP
       using disown -h.
```

regular:
> Re: However, when the test process is interrupted with ^C, sbot survives. I don't know why that is. 

I still don't understand this one however. If `^C` kills the test process, the test process closes it's file handles, which in turns should unlock the `read` call in the shell script that wraps sbot, and then goe son and kill sbot.  The only reason I can think of why this doesn't happen is, that in this scenario, the shell script is killed before the read unblocks, which is weird, because if this was the case, the whole thing would not have been necessary in the first place ...

PR #560, opened 2018-10-18
WIP: Add hacky "test" for new meta code

christianbundy:
This adds and `addMap()` to test https://github.com/ssbc/secure-scuttlebutt/pull/228/ and see what breaks. Currently everything seems to be working great. It's clear to me that this test shouldn't be in index.js, but is there somewhere else it should be placed? I'm hesitant to put it in its own isolated file because it affects the output of *all* tests, but I suppose maybe I could just add it as a test-only plugin? Not sure what the right call is here.

### DIFF
--- a/index.js
+++ b/index.js
@@ -171,11 +171,22 @@ function createSbot() {
     appKey: require('./lib/ssb-cap')
   })
     .use(SSB)
+    .use(api => {
+      api.addMap((msg, cb) => {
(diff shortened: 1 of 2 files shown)


christianbundy:
cc: @dominictarr @pietgeursen How should code like this be tested? It seems useful to add this to *all* tests, but alternatively I could just do a single test for this.

PR #562, opened 2018-10-24
Set `try_often(10, ...)` for all instances

christianbundy:
I'm hoping that this is the reason our Travis builds are failing.

@regular @cryptix Is there any reason to keep this at 3?

---

Nope, didn't work.


christianbundy:
Computers are *so fast* except when they're not.

If it was meant to be 10 I'll leave this open and we can merge whenever. Any chance you know why the test keeps failing on Travis? It looks like it's the 127.0.0.1 test, too.

PR #563, opened 2018-10-24
Bugfix and cleanup in test/bin.js

regular:
I simplified the test code and fixed at least one bug in it. This could heal travis.


regular:
This seems to have fixed it (or we were just lucky, because the machine is under less load now)

It's fun to watch travis in realtime, didn't know!

PR #569, opened 2018-11-01
support multiple scopes

dominictarr:
mainly this is updating the invite plugin, so that if allowPrivate is used, it uses the local scope instead of public.

depends on: 
* https://github.com/ssbc/multiserver/pull/32
* https://github.com/ssbc/ssb-config/pull/25
* https://github.com/ssbc/secret-stack/pull/31
* and ssb-ws@3.0.2 @regular I couldn't quite figure out what was going on with the multiple servers thing on ssb-ws@4

all the tests are now passing again, and `getAddress <scope>` returns something sensible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,4 @@ node_modules
 sbot.js
 .ssbrc
 *.log
-
-
-
+coverage
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "minimist": "^1.1.3",
     "mkdirp": "~0.5.0",
     "multiblob": "^1.13.0",
-    "multiserver": "^1.13.4",
+    "multiserver": "^2.0.0",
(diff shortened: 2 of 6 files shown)


regular:
@dominictarr what's your problem with the multiple ws servers thing? If people chose to have multiple ws incomings, it should just work right? Before it was a singleton.

dominictarr:
well, sorry what I mean is that when I got this working it was on the v3 branch, and then when I realized that I didn't have time to update it to master, and wasn't entirely sure how it worked. so I left it. It was really just a time constraint thing.

hmm, you are right, though it is a singleton... I think the solution is to move createServer inside sbot.multiserver.transport({create: function () {...}}) though, that will be cleaner than making a keyvalue map of the servers by host/port

PR #573, opened 2018-11-11
Remove ssb-links

arj03:
I was looking at the various flume indexes and bumped into the links and links2. While links is used internally by secure-scuttlebutt for things like messagesByType, links2 doesn't really seem to be used at all. Maybe it was used for some of the old functions that was removed once?

Doing a full search on github reveals that it is exposed in ssb-ws [here](https://github.com/ssbc/ssb-ws/blob/master/index.js#L28) and used in patchfoo. But patchfoo installs it as a plugin, so that shouldn't be any problem. I then checked mvd at it appears only to be referenced in [manifest](https://github.com/evbogue/mvd/blob/392e50fcc67f4ad4b873fc95b096e3018404f255/manifest.json#L70) but never used. @evbogue can you confirm?

Ping @regular are you using that api for anything?

The thing is that this index was the biggest of all on my machine. 233M. So it would be really nice not having to maintain that index if its not really used anymore. And if it is, then people should include it themselves. Also scuttle-shell includes this [exactly](https://github.com/ssbc/scuttle-shell/blob/master/server.js#L59) for patch-foo, so when people start using it this shouldn't be a problem. On the other hand it would probably be a good idea to see if patchfoo could use another api as to remove the indexing overhead in there. But that is totally unrelated to this PR imo, just a friendly reminder :)

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -57,6 +57,13 @@ if (argv[0] == 'server') {
   // add third-party plugins
   require('./plugins/plugins').loadUserPlugins(createSbot, config)
 
+  if (argv[1] != '--disable-ssb-links') {
+    if (!createSbot.plugins.find(p => p.name == 'links2')) {
+      console.log("WARNING-DEPRECATION: ssb-links not installed as a plugin. If you are using git-ssb, ssb-npm or patchfoo please consider installing it")
+      createSbot.use(require('ssb-links'))
+    }
+  }
+
   // start server
 
   config.keys = keys
--- a/bin.js
+++ b/bin.js
@@ -52,7 +52,6 @@ if (argv[0] == 'server') {
     .use(require('./plugins/local'))
     .use(require('./plugins/logging'))
     .use(require('ssb-query'))
-    .use(require('ssb-links'))
     .use(require('ssb-ws'))
     .use(require('ssb-ebt'))
   // add third-party plugins
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "ssb-ebt": "^5.1.4",
     "ssb-friends": "^3.1.3",
     "ssb-keys": "^7.1.1",
-    "ssb-links": "^3.0.2",
     "ssb-query": "^2.1.0",
     "ssb-ref": "^2.13.3",
     "ssb-ws": "^3.0.1",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "ssb-ebt": "^5.1.4",
     "ssb-friends": "^3.1.3",
     "ssb-keys": "^7.1.1",
+    "ssb-links": "^3.0.2",
     "ssb-query": "^2.1.0",
     "ssb-ref": "^2.13.3",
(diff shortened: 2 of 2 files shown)


cryptix:
git-ssb and patchfoo make use of it AFAIK.

arj03:
Wait a minute, isn't patchfoo only using [links](https://github.com/ssbc/patchfoo/blob/9cdce1f0fa8e7969eb1ab545e716b4630ae1ec68/lib/app.js#L86) and not links2? I think I found an old readme.

evbogue:
@arj03 ssb-links is only used in http://github.com/evbogue/mvd via ssb-avatar. If we can modify ssb-avatar to use ssb-query then I can remove it from mvd.

evbogue:
Oh, but that's links and not links2. Yes, I don't use links2 in mvd. 

I was looking at the manifest file yesterday and I thought to myself that I should clean that up.

arj03:
Excellent. Thanks @evbogue for the confirmation!

Also @cryptix is right about git-ssb. It's used in the name command only. I'll see if I can rewrite it to use backlinks instead.

Yes, lazy indexing would be excellent! Hmm that is a good question wtf backwords compatibility @regular. Thanks for bringing it up. It's not like it's impossible to use ssb-links with it removed from default install. You just have to install it as a plugin instead.

arj03:
I reached out to cel at it appears to be used both in ssb-npm and patchfoo (in a version not on github). 

The thing is, the plugin is no longer used in sbot, so it shouldn't be there the same way as backlinks is not part of core. But I don't want to break stuff for existing users. Dynamically loading a plugin has the problem that this means the plugin first starts indexing on first usage, which is not ideal and also quite counter-intuitive that ssb-links should behave different than other flume plugins. *sigh*

regular:
@arj03 How about an explicit opt-out for people that really don't can afford those ~250MB storage space? So, by default everything stays as-is, and users can explicitly disable built-in plugins if they know what they are doing?

arj03:
Sure that would be an option. And maybe give a warning if users don't have ssb-links installed as a plugin. That way we can deprecate the option once people have install ssb-links as a plugin. I don't really like extra complexity.

I'd be really interested to hear how this together with the changes in secure-scuttlebutt changes initial sync. Any chance you can test that @staltz?

arj03:
@staltz Yes it should. This index is quite big and also secure-scuttlebutt@18.5 has improvements that makes write speed 10-15% faster.

staltz:
@arj03 I was excited to try the new versions, but ended up postponing because it may take a bit more work, there are some breaking changes to adapt to (ssb-friends v3 at least). See https://gitlab.com/staltz/manyverse/issues/239

arj03:
Should be ready now. To sum up:

- By default sbot server will now warn about deprecation of ssb-links
- If ssb-links is installed as plugin (config needs to be "ssb-links": "links2") all is good
- ssb-links can be disabled with --disable-ssb-links

PR #580, opened 2018-11-27
Don't nuke config if its not valid json.

arj03:
The problem is that earlier if one had invalid json it would just end up with an empty dict and write the plugin to that, thus nuking your carefully handwritten config. This bails instead and actually shows what is "wrong" with the file.


dominictarr:
gossip.json is different, because it's not intended to be hand-edited. :+1: on a merge

PR #585, opened 2018-12-05
Fix config file read

### LINKED ISSUE #584: Error: ENOENT: no such file or directory, open '/tmp/test-plugins1/config'
Failing with `npm test`:

```
Error: ENOENT: no such file or directory, open '/tmp/test-plugins1/config'
    at Object.openSync (fs.js:436:3)
    at Object.readFileSync (fs.js:341:35)
    at writePluginConfig (/home/christianbundy/src/ssbc/scuttlebot/plugins/plugins.js:64:42)
    at /home/christianbundy/src/ssbc/scuttlebot/plugins/plugins.js:132:19
    at /home/christianbundy/src/ssbc/scuttlebot/node_modules/mv/index.js:34:26
    at FSReqWrap.oncomplete (fs.js:141:20)
npm ERR! Test failed.  See above for more details.
``

christianbundy:
This was recently refactored to prevent the server from overwriting invalid config files, but had a bug where it would crash when the config file was missing. This commit aims to be the best of both worlds, where we can continue safely on missing config files but crash loudly when the config file has the wrong permissions or is invalid JSON.

Resolves #584

### DIFF
--- a/plugins/plugins.js
+++ b/plugins/plugins.js
@@ -61,14 +61,28 @@ module.exports = {
     function writePluginConfig (pluginName, value) {
       var cfgPath = path.join(config.path, 'config')
       // load ~/.ssb/config
-      var existingConfig = JSON.parse(fs.readFileSync(cfgPath, 'utf-8'))
+      let existingConfig
+      fs.readFile(cfgPath, 'utf-8', (err, data) => {
+        if (err) {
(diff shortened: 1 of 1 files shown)


arj03:
Looks good to me.

christianbundy:
Thanks!

Issue #579, opened 2018-11-26
Gossip has stopped working


christianbundy:
I'm having an issue where gossip isn't happening no matter how long I leave Scuttlebot running. On 13.2.0 I was seeing #578, but even after downgrading to 13.0.3 the output of `sbot status` shows `gossip: {}` and I haven't been able to receive any messages for about 9 hours. I've tried ^12.0.0 and it said "Error: listen EADDRINUSE :::8989", and ^11.0.0 had the same behavior as 13.2.0 and 13.0.3. Weird.

Has anyone else been having problems? I'm seeing issues with my global `sbot`, Patchbay, and Patchwork. I've tried restarting my machine but it doesn't seem to fix anything. These are the software upgrades I did today, just in case one of these may be misbehaving:

```
[2018-11-25 09:35] [PACMAN] Running 'pacman -Syu'
[2018-11-25 09:35] [PACMAN] synchronizing package lists
[2018-11-25 09:35] [PACMAN] starting full system upgrade
[2018-11-25 09:36] [ALPM] transaction started
[2018-11-25 09:36] [ALPM] upgraded bind-tools (9.13.3-3 -> 9.13.4-1)
[2018-11-25 09:36] [ALPM] upgraded bind (9.13.3-3 -> 9.13.4-1)
[2018-11-25 09:36] [ALPM] upgraded libwmf (0.2.8.4-14 -> 0.2.10-1)
[2018-11-25 09:36] [ALPM] upgraded v4l-utils (1.16.1-1 -> 1.16.2-1)
[2018-11-25 09:36] [ALPM] transaction completed
[2018-11-25 09:36] [ALPM] running 'gdk-pixbuf-query-loaders.hook'...
[2018-11-25 09:36] [ALPM] running 'gtk-update-icon-cache.hook'...
[2018-11-25 09:36] [ALPM] running 'systemd-daemon-reload.hook'...
[2018-11-25 09:36] [ALPM] running 'systemd-sysusers.hook'...
[2018-11-25 09:36] [ALPM] running 'systemd-tmpfiles.hook'...
[2018-11-25 09:36] [ALPM] running 'systemd-udev-reload.hook'...
[2018-11-25 09:36] [ALPM] running 'systemd-update.hook'...
[2018-11-25 09:36] [ALPM] running 'update-desktop-database.hook'...
[2018-11-25 17:06] [PACKAGEKIT] synchronizing package lists
[...]

PR #592, opened 2018-12-12
some fixes to gossip for v13

dominictarr:
Hey I think this fixes the problems we've been having.

should fix https://github.com/ssbc/scuttlebot/issues/579

main fix was https://github.com/ssbc/ssb-ref/pull/30 which allows sbot to rebuild the peers list from pub messages. @christianbundy @mmckegg @arj03 @mixmix does this fix problems you were seeing?

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "ssb-keys": "^7.1.1",
     "ssb-links": "^3.0.2",
     "ssb-query": "^2.1.0",
-    "ssb-ref": "^2.13.3",
+    "ssb-ref": "^2.13.9",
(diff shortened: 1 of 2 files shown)


dominictarr:
I don't think the format needs to change. I'm not sure how we ended up loading the gossip.json file in two places, prehaps someone moved it in their PR and then someone updated it and it merged both back into different places? anyway, the place that remains now adds peers in a way that will fix the format.

dominictarr:
added a commit to ensure that the output gossip.json is something that earlier versions of scuttlebot will understand.

PR #594, opened 2018-12-13
added db_prefix to namespace indexes for each app so they don't smoosh eachother

dangerousbeans:


mmckegg:
@dangerousbeans In my opinion, this loses the advantage of having shared indexes at all.

What about instead of a namespace, we prefix the index version?

mmckegg:
Oh sorry @dangerousbeans, just caught up with the thread on ssb about this.

Ah, @dominictarr recommend you do it this way. Hmm, I think that your initial idea about version number is much more useful, but that it should be a config option (if someone just uses Patchwork, there is no advantage to keeping old versions around).

index.js
```diff
@@ -101,6 +104,7 @@ var SSB = {
       //temporary!
       _flumeUse                :
         function (name, flumeview) {
+          name = opts.db_prefix + name
```
christianbundy: I might suggest:

```suggestion
          name = opts.dbPrefix + name
```

dangerousbeans: 👍

dominictarr:
@dangerousbeans please try and run this before you make the PR, and you'll see what the problem is ;)

dangerousbeans:
@dominictarr it's running - as far as I can tell it works alright?

dominictarr:
oh you fixed it. @dangerousbeans if you put a / in it will put everything inside a directory, much tidier. `prefix ? path.join(prefix, name) : name` also, please make the option camelCase. There are some snake_case things, but only internal variables and c bindings. please keep the api consistent. or maybe it should be `{flume: {prefix: dbPrefix}}` we havn't needed another flume config, but we should probably keep them together.

dangerousbeans:
@dominictarr should we just namespace via the npm version of flume, or whatever package gets upset? (is that available in scope somewhere?)

christianbundy:
Is it just me, or will this have very strange behavior when `opts.dbName` is undefined?

```javascript
const opts = {}
const name = 'patchbutt'
opts.dbName + name // => "undefinedpatchbutt"

christianbundy:
Hasn't been active for over a year, I'm going to close. Please re-open if you'd like to see this merged!

PR #599, opened 2018-12-21
add config.gossip.autoPopulate to allow disable of auto build of peer table

mmckegg:
It is currently not possible to control which peers get added to the gossip table because this is generated deep inside the gossip plugin.

**This PR adds the config option `gossip.autoPopulate` that when set to false, prevents the peer table from automatically being populated from `{type: 'pub'}` messages, allowing the instance to manually populate however they want.**

An example of where this is needed is if you want to exclude adding blocked pubs to the peer table.


dominictarr:
heh, just had a branch lying around for this! http://github.com/ssbc/ssb-device-address is relavant if you havn't seen it yet!

PR #604, opened 2018-12-23
Add friends config option to connect to peers marked as friends

arj03:
This plays well with [ssb-friend-pub](https://github.com/ssbc/ssb-friend-pub) module. 

I reused the "friends" type for adding these peers so that friend-pub also should work together with the global module and prioritize friends.

The idea with this flag is that one should be able to only connect to peers marked as friends.


dominictarr:
looks good!

parsa2000200:
- [ ] ji@dominictarr

PR #610, opened 2019-01-05
refactor all plugins out, but still run their tests

dominictarr:
This is a start at the idea I suggested here: https://github.com/ssbc/ssb-server/issues/602#issuecomment-449681779

There is still some more work to do:
* still more plugins to be refactored out (but I got the big ones)
* ssb-db should be a plugin so that you do `SecretStack().use(require('ssb-db').use(...)` etc

I just wrote a quick bash script (test/deps.sh) to run the tests for the deps. if you are maintaining a ssb-server distribution, you need to manually maintain the dev deps. (you could write a script to update this though). Also, if one dep's test fails, it's hard to see which. It would be good to organinize that better, or have a way to run a single test from a single dep, but this gets us started.


staltz:
Really support this PR getting merged, thanks Dominic

dominictarr:
Okay this is now ready to merge. now ssb-server is _just a distribution_.
In this pattern, when testing, plugins should do `SecretStack({}).use(require('ssb-db').use(...)` instead of `SsbServer.use(...)`

but this means there will be only a few lines of code in ssb-server, so it can easily be copied as a template for alternative distributions of sbot. also, it will be perfect with a shrinkwrap. most importantly: by running the tests of the most important deps, we can verify that all the deps are compatible with each other and everything in that distribution will work, before releasing a new version.

Is everyone cool with that? @arj03 @clehner @christianbundy @mixmix @mmckegg @staltz @regular

dominictarr:
oh yes, depends on: https://github.com/ssbc/ssb-db/pull/247

regular:
@dominictarr I like it a lot!

staltz:
Totally support this. We could also consider at some point renaming this repo to be something like "ssb-standard-server" since it would be *one of the distributions*.

christianbundy:
I like this too. My favorite pattern here is how ESLint manages [eslint:recommended](https://eslint.org/docs/rules/). It's a recommended set of plugins that most people start with, but they don't make any assertions that it's the "correct" set of plugins.

mixmix:
This is pretty exciting and relieving

Could this put us on the path to ssb-cli being an external plugin too?

mixmix:
Oh I would say can we stabilise after getting https://github.com/ssbc/ssb-server/pull/615 merged before making this major jump?
I'm keen to get all my `ssb-config` work merged and into patchbay and successfully building installers on all platforms. I think the setup for Windows is currently broken. Getting these installers working can be a major time sync.

dominictarr:
@mixmix yes, ssb-cli makes a lot more sense after this, because the same one could talk to any ssb-server distribution.

dominictarr:
this is not _quite_ ready, it turns out, when all the deps are directly installed I am getting an error
hmm it might be a race... will look at this again tomorrow

dominictarr:
okay this is mergable now, but I want to merge https://github.com/ssbc/ssb-config/pull/31 and https://github.com/ssbc/ssb-server/pull/613 along with it.

dominictarr:
oops, I messed that up, but everything fixed in 14.0.1

PR #612, opened 2019-01-08
add a useable example

mixmix:
I recently tried to modifying how I start my `ssb-server` and I couldn't find how or where to learn about configuration.

This PR adds:
- a clear link out to `ssb-config` (where the config you need to know about is to start the server!)
- clear link to other modules / docs
- a complete `example use` code snippet
  - the prior example was junk <3

---

_I had a bunch of kinda wound up feelings writing this. I imagined a new dev arriving at this README and just feeling stupid. I felt stupid and I'm an "experienced" dev in this system. So please, in reading this, it might not be perfect but lets make some improvement that throws people a bone (idiom which means help people just a little bit)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,15 +36,13 @@ There are already several applications built on ssb-server,
 one of the best ways to learn about secure-scuttlebutt is to poke around in these applications.
 
 * [patchwork](http://github.com/ssbc/patchwork) is a discussion platform that we use to anything and everything concerning ssb and decentralization.
-* [patchbay](http://github.com/dominictarr/patchbay) is another take on patchwork - it's compatible, less polished, but more modular. The main goal of patchbay is to be very easy to add features to.
+* [patchbay](http://github.com/ssbc/patchbay) is another take on patchwork - it's compatible, less polished, but more modular. The main goal of patchbay is to be very easy to add features to.
 * [git-ssb](https://github.com/clehner/git-ssb) is git (& github!) on top of secure-scuttlebutt. Although we still keep our repos on github, primary development is via git-ssb.
 
 it is recommended to get started with patchwork, and then look into git-ssb and patchbay.
 
-## Example Usage
+## Example Usage (bash)
 ```bash
-# In bash:
-
 # Start the server with extra log detail
 # Leave this running in its own terminal/window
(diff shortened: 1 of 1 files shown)


mixmix:
```
  .use(require('ssb-server/plugins/replicate'))
  .use(require('ssb-server/plugins/invite'))
  .use(require('ssb-server/plugins/local'))
  .use(require('ssb-server/plugins/logging'))
  .use(require('ssb-server/plugins/unix-socket'))
.use(require('ssb-server/plugins/no-auth'))
```

here's a thing I'm not sure about @christianbundy ... what's the current minimum set of internal plugins you need to include to not explode? .... I remember when I tried figuring this out a few months ago there were 2-3 which seemed non-negotiable (ironically) ... but since then sbot has change quite a bit....

README.md
```diff
@@ -61,19 +59,45 @@ ssb-server log
...
 ```
+## Example Usage (js)
+
 ```js
-// In javascript:
+var Server = require('ssb-server')
+var config = require('ssb-config')
+var fs = require('fs')
+var path = require('path')
+
```
christianbundy: Just tried to run this example and it threw an error about my lack of keys. I had to add this:

```js
config.keys = require('ssb-keys').loadOrCreateSync(path.join(config.path, 'secret'))
```

christianbundy:
I'm not sure. I just tried testing and I couldn't get it to explode, even without any plugins. Things like *get* don't work, but I think that's expected (?) since there weren't any plugins loaded to handle it.

README.md
mixmix: right, nice catch! (I wish JS had that rust thing where example code gets run as a real test!)

I have personal feeling that should be in `ssb-config` yeah?
As in, you have to do this _every time_ right, and that things job is to get the default shit together for the server yeah?

christianbundy: I hadn't really considered that, but yeah, that sounds exactly like something that should be declared in ssb-config.

mixmix: here's the PR : https://github.com/ssbc/ssb-config/pull/29

mixmix:
Ok I'm gonna assume that other PR with ssb-config is merged ... at least this is better than what's currently there.

PR #615, opened 2019-01-14
fix invite tests when device has public ipv6 address

dominictarr:
I think this should get things passing for you @mixmix can you run the tests on this branch?

### DIFF
--- a/test/invite.js
+++ b/test/invite.js
@@ -109,9 +109,13 @@ tape('test invite.accept doesnt follow if already followed', function (t) {
           if(err) throw err
           t.equal(ary.length, 1)
 
+          //it's rare, but some times, someone's home computer has a public address.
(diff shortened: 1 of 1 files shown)


dominictarr:
you might need https://github.com/ssbc/ssb-config/pull/25 too

mixmix:
can confirm this passes all test : 
- by itself
- with  ssbc/ssb-config#25

mixmix:
Back on the wifi at home:
- :x: by itself
- :heavy_check_mark: with ssbc/ssb-config#25

dominictarr:
cool. I didn't expect this to run on it's own on your weird home setup mix

dominictarr:
touche

dominictarr:
I have merged this into 13.6.1 (also depending on ssb-config@3) since it's just a fix.

PR #617, opened 2019-01-18
Fix lockfile

christianbundy:
It looks like the current `npm-shrinkwrap.json` is invalid:

```
Invalid: lock file's ssb-config@3.2.0 does not satisfy ssb-config@^3.2.2
```

This branch resolves the issue.

---

This branch also removes unused dependencies!


christianbundy:
I've pulled in the latest changes from master and they have the same problem:

```
npm ERR! Invalid: lock file's secret-stack@5.1.0 does not satisfy secret-stack@^6.0.1
npm ERR! Invalid: lock file's ssb-db@18.6.5 does not satisfy ssb-db@^19.0.0
npm ERR! Missing: ssb-gossip@^1.0.3
npm ERR! Missing: ssb-invite@^2.0.3
npm ERR! Missing: ssb-replicate@^1.0.4
npm ERR! Missing: interleavings@^0.3.1
npm ERR! Missing: pull-bitflipper@^0.1.0
npm ERR! Missing: run-series@^1.1.8
npm ERR! Missing: ssb-generate@^1.0.1
```

I've pushed another commit resolving these as well. For what it's worth, I'm running `npm ci` to test the current lockfile and running `npm install` if it fails, then I'm adding the new lockfile to a commit and pushing.

test/deps.sh
```diff
@@ -9,6 +9,7 @@ set -e
 test () {
   echo "## TESTING DEPENDENCY: $1"
   pushd node_modules/$1
+  npm install --only=dev
```
christianbundy: I've added this line to ensure that we have the correct devdeps for running tests.

dominictarr:
I couldn't figure out how you got that error for a while... my general complaint about npm not showing wether you are using the shrinkwrap or what

dominictarr:
you were running `npm ci` ? I managed to get that error with `npm ls --package-lock`.
I think npm ci still looks at the package.json though... I wish there was a way to just install from a lone shrinkwrap, so that I could just email you the dep tree I'm debugging.

christianbundy:
> I think `npm ci` still looks at the `package.json` though... I wish there was a way to just install from a lone shrinkwrap, so that I could just email you the dep tree I'm debugging.

Sweet,that's what `npm ci` does. It looks at both `package.json` and `package-lock.json`, ensures that `package-lock.json` satisfies the dependencies in `package.json`, and installs the exact dependency tree described by `package-lock.json`.

dominictarr:
because then you can make tools for manipulating the dep tree, that output a shrinkwrap, and npm can install it. A while a go, I was trying this, changing the resolved url (to local blobs url) hoping that npm would install that, provided that it matched the integrity, but it doesn't! It rerequests it from the registry!)
I ended up making this: https://github.com/dominictarr/npm-install-shrinkwrap

mixmix:
I did the dependancy removing part here : https://github.com/ssbc/ssb-server/pull/622/files 

, will merge master in so we can see what's left

test/deps.sh
```diff
@@ -17,6 +17,7 @@ name () {
...
   npm test | name $1
```
mixmix: I wasn't use how to merge this line.
Is currently:

```bash
npm test | name $1
```

On this branch is was originally: 
```bash
npm test
```

christianbundy:
Hey @mixmix, sorry to waste your time on this. I saw that Dominic opened a counter-PR (#618) so that got merged instead. I'd still rather handle sub-dependencies automatically rather than hardcoding them, but I suppose I'll do that in a new PR. Thanks for the review :heart:

mixmix:
oh weird ... I wonder why the master branch still had those changes ... I hope I didn't mess something up

dominictarr:
@christianbundy I'm not against a script that updates the dev deps, to include the dev deps of the plugins. The important thing is that the dev deps are in the lock file.

christianbundy:
@dominictarr Could we discuss on #623 instead?

dominictarr:
yes of course

PR #623, opened 2019-01-28
Add missing dependencies, dynamically for child deps

christianbundy:
This commit makes two changes: the least controversial change is to add
two modules to `package.json` which are required in our test files.

The second change is to add the necessary devDependencies for child
dependencies when we run their tests. Specifically, this means that
we're installing all `devDependencies` except those that have been
included as production dependencies in the ssb-server package.

This was originally added in #617, but when #618 was merged it hardcoded
these dependencies instead of adding them dynamically. A few commits
later these hardcoded child devdeps were removed in 95785cb, which means
that `npm test` started failing again. This commit resolves the problem.


christianbundy:
I've added a patched version of ssb-blobs to resolve #624, basically it just replaces `console.log()` with `debug()` so that it doesn't output tens of thousands of lines of output by default.

christianbundy:
From https://github.com/ssbc/ssb-server/pull/617#issuecomment-458711509

> @christianbundy I'm not against a script that updates the dev deps, to include the dev deps of the plugins. The important thing is that the dev deps are in the lock file.

@dominictarr I totally agree, as long as we're talking about the plugins' lockfiles. Is that what you mean?

dominictarr:
we can't merge this - dynamic dev deps will break test determinism. what we could have is a script that merges the dev deps of the plugins (that will be tested) into the ssb-server's `devDependencies` but we should not just install them like this. same problem as I expressed here: https://github.com/ssbc/ssb-server/pull/618#issuecomment-455902772

as it seems I'm not being sufficiently clear

```
ssb-friends -> ssb-gossip (old version)
ssb-server -> ssb-friends, ssb-gossip (new version)
```
when you check out ssb-friends npm will install the version resolved looking at the dev deps - the old version. the tests will pass.
when you clone ssb-server, and do npm install it will install ssb-friends and ssb-gossip, but then when you do the dynamic install of ssb-friends' deps, npm will think it's has resolved the wrong version of ssb-gossip for ssb-friends, so it will install the old version in `ssb-friends/node_modules/ssb-gossip`. 
But lets suppose that ssb-friends would fail if it runs with new `ssb-gossip`, since this is the version it will be run with in the new `ssb-server` distribution, we need to actually run the tests with that version.
If we dynamically install the deps for `ssb-friends` it will get the old version and that will cause `ssb-friends'` tests to pass, when actually they should fail, as they ought to run with the same version of `ssb-gossip` as `ssb-server` is using.

Hence, we cannot use dynamically installed deps. Also, we could just get indeterminism from not using the lock file. If we are gonna use lockfiles, we should use it for everything.

test/deps.sh
```diff
@@ -17,6 +19,11 @@ name () {
 test () {
   echo "## TESTING DEPENDENCY: $1"
   pushd node_modules/$1
+  npm install --only=dev
+  pushd node_modules
+  # remove duplicates of upstream deps
```
christianbundy: @dominictarr 

Have you already looked at the code? I think this resolves the problem you've illustrated, but if it doesn't I'd love to understand why it won't work.

dominictarr: oh I was reacting to "dynamic" and didn't see the "# remove duplicates of upstream deps" comment.
I must say I don't really understand how it works... it looks more like it deletes the production after each test run?

Oh so it gets the production deps used by `ssb-server`, then it cd's into a dev dep, then it installs the dev deps, then it deletes anything which _also_ happens to be a product dep of `ssb-server`, so only extra needed dev deps remain. Hmm, so I think this does address my concern that duplicate deps will be created...

but I honestly think this is not as nice as just putting those extra deps into `ssb-server`'s devdeps.
You campaigned for a long time to get ssb-server to use a lockfile, now it does - but this abandons the lockfile for the dev deps! If we are gonna do a lock file _everything should be in the lockfile_

dominictarr: how do you feel about a script that automatically includes them in the dev deps?
I am a lockfile _all-or-nothing-ist_

sure, the dev deps appearing unused is an easy mistake, but that's just needs a little documentation about how the integration tests work.

dominictarr: btw, the dev deps are the things needed to do development - i.e. run the tests. because of running integration tests, they _are_ also our dev deps.

christianbundy: We already have a script that specifies our dependencies in a lockfile: `npm install`.

If you're interested in specifying plugin devdeps in a lockfile then I'd propose that we add a lockfile to the plugin repos. This would resolve your concern *and* make those repos more conventional, which increases accessibility.

Would you be open to pull requests adding lockfile to each of the plugin repos?

dominictarr: @christianbundy I suggested a compromise that I thought would work for both of use and asked how you felt about it - please work with my by at least describing what your problems are with it, so as to show me openings towards something else that might work.

I do not agree that lockfiles to each of those projects is the right solution. The point is to test those modules with `ssb-server`'s deps, not with the deps specified when `ssb-gossip` or what have you's deps. (the plugins deps are probably old, I don't want to go and publish new versions of all the plugins just because one has been updated)

This is a weird move, but it's also about accessability. I want the test suite to have very good coverage so that you can be confidant that the PR etc you've made doesn't break anything. (people are afraid to break stuff, including me)

dominictarr:
doing a dynamic install during testing is also as unconventional as adding seemingly unused dev deps (but may break offline testing). but I think static devdeps are the lesser evil as it's deterministic.

christianbundy:
Maybe we could add your proposed script as a [postinstall](https://docs.npmjs.com/misc/scripts#examples)? This would give us the benefit of

dominictarr:
update: turns out a bash mistake fixed in 122a5eabef42454b2d86da9b0be485f9f3235827 was swallowing false positives. If that had been in place @mixmix wouldn't have removed those deps because the tests would have broken ... currently tidying up my fix and will PR.

@christianbundy yes a post install script would work

christianbundy:
I originally closed this because commits were pushed straight to `master`, but it looks like https://github.com/ssbc/ssb-server/commit/a951355b3518671940d89ec1bfafa1310829f61f and its ancestors have been unpublished, so this is now good to go again.

dominictarr:
allow me to explain that the intention of pushing those was only to get what was already there working again, not to squash this on going discussion. Once I realized it was more complicated than i originally thought I reverted and made it a PR because I wanted there to be a record of how and why

christianbundy:
No problem, I didn't think there was any malice. :heart: I appreciate the follow-up anyway, and I'm happy we have a fix in `master` now. Closing!

PR #626, opened 2019-01-30
Fix integration tests

dominictarr:
A few days ago @mixmix [removed dev deps](https://github.com/ssbc/ssb-server/commit/95785cb598043c1b820f3ef6cf708a579585de36) which were needed to run the integration tests, but strangely it managed to publish (failing integration tests should have prevented the publish)

Turns out integration tests were ignoring failures, because of the thing I added to show what integration test they were in. It was swallowing the error code. [fixed that by ensuring the name command preserves the exit code](https://github.com/ssbc/ssb-server/commit/87e00e5a6965f329569682c0fd4a2ff5eca2d1bf)

To make the integration tests load ssb-server and run their tests, they need to find the current ssb-server repo. npm modules cannot refer to them selves in the 3rd person, so `require("module_name")` only works if you are not "module_name", this is a pity, and means we need to add a symlink
`cd node_modules; ln -s ../ ssb-server` to get it to work. _however_ this interacts with `npm ls` so we need to make sure we _remove_ that symlink before doing `npm ls` otherwise it will likely report many missing modules (these aren't really missing and the tests will pass)

@christianbundy I think this covers the other quirks needed to run the integration tests in this way


christianbundy:
LGTM, merging now.

dominictarr:
@christianbundy btw, when you merge can you publish? (and comment the new version number?)

PR #628, opened 2019-01-31
Add pretest

### LINKED ISSUE #625: How should we specify devDependencies for integration tests?
I don't think this is the global optima, but as a local optima for cross-module integration tests I'd be open to using your idea as a postinstall script that gathers plugin devDependencies and adds them to our lockfile. This has three benefits:

1. Increased determinism for our testing.
2. It's automatic and always remains up-to-date.
3. Contributors don't have to learn anything new.

cc: @dominictarr (I've closed #623 because you pushed your commits straight to `master` instead)

christianbundy:
@dominictarr 

If we're going to do it then we may as well automate it. Here's a first stab that runs before `npm test`:

1. Collect the list of dependencies from ssb-server's `package.json`
2. Collect the list of dependencies from each plugin
3. If a plugin devDependency is incompatible with ssb-server or another plugin, throw an error
4. If we have new devDependencies to install, install and exit 1 (to ensure bad commits don't pass test)

I think this is better than my previous suggestion of `npm install` because it was running after `npm ci` and messing with our determinism.

Resolves #625

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -465,6 +465,7 @@
       "requires": {
         "anymatch": "^1.3.0",
         "async-each": "^1.0.0",
+        "fsevents": "^1.0.0",
         "glob-parent": "^2.0.0",
         "inherits": "^2.0.1",
         "is-binary-path": "^1.0.0",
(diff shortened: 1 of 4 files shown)


dominictarr:
I like it!

Questions (I'll try to answer them my self):

* when two plugins have a dev dep that conflicts, the pretest script fails. a@2 a@3 would fail, it this case you will need to fix the plugins.
* intersecting ranges work `a@^2.1` `a@^2.2` would pass. If that particular dev dep is not available, it will install it, and the local package.json will be updated, with a ^ range of the latest version of that module (npm default). It is strongly advised to use ^ ranges. obviously if two plugins have the same range on a dep that is considered an intersect.
* if an install is necessary, the pretest will exit 1. This will cause the tests not to run, but if you run the test again it should work (if the tests pass, now that deps are present). This is so that a running test in `prepublishOnly` script prevents publishing broken code.

dominictarr:
merging into 14.0.4

note: since the point here is that anyone publishing a ssb-server distribution should also run the integration tests, we want it to be as easy as possible. Theirfore we'll need to publish this stuff as modules, so that they can easily import it. (copying the scripts will be maintainance nightmare) on that note still need to publish the other plugins/

PR #635, opened 2019-02-07
script for test coverage

dominictarr:
test coverage for integration tests, and all the important modules: `ssb-*,flume*,secret-stack,muxrpc,secret-handshake`

currently coverage is 73%!

I think the best way to improve it is to kill things that arn't actually being used.
and then to write a few more tests.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4,6 +4,144 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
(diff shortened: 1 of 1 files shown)


dominictarr:
because the glob script for things to include doesn't support exclusions very well, duplicate deps kept appearing in the coverage output, so I ended up updating everything so that all got tidied away.

ahdinosaur:
is this pull request still a thing? the only changes i see are to the shrinkwrap.

dominictarr:
there is a `coverage` script, so must have merged those commits.

PR #642, opened 2019-03-02
Fix shrinkwrap file

christianbundy:
I didn't spend any time debugging the problem, this is just the result of `npm run reinstall`.

Resolves https://github.com/ssbc/ssb-server/issues/641


christianbundy:
I'm going to self-merge this as it's trivial and currently `npm install ssb-server` fails and that's very bad.

PR #644, opened 2019-03-04
Update npm dependencies

christianbundy:
This bumps some dependencies so that all of the integration tests pass again (:tada:).

Should the new ssb-ws changes be pulled in? If so, I imagine that would be a major version bump to be released as ssb-server@15.0.0.


christianbundy:
Maybe I should just be pushing this stuff to `master`? :upside_down_face:

PR #652, opened 2019-03-31
reinstate --help

dominictarr:
pretty much a complete rewrite of the internal help system.

now with muxrpcli@3 has each plugin own it's own help - plugins should have a `help` method that returns a data structure describing all commands and all their arguments. That has been implemented for ssb-db and ssb-gossip so far, but all the plugins will need that.

This data is also machine readable, which I think will lend to other useful things, such as validation of inputs. Also since it describes the patterns, it would be usable for other types of interface, http, cli or javascript.


dominictarr:
modules that need "help"
* [] ssb-replicate
* [] ssb-friends
* [] ssb-blobs
* [] ssb-invite
* [] ssb-peer-invites
* [] ssb-query
* [] ssb-ooo
* [] ssb-device-address
* [] ssb-search

staltz:
Could the *help* data structure be described somewhere as a convention? It smells like a module in itself (similar to multiserver-address).

dominictarr:
@staltz yes, it's documented here: https://github.com/dominictarr/muxrpc-usage#data-structure

dominictarr:
none of the plugins moved in https://github.com/ssbc/ssb-server/pull/664/files have commands, so merging that doesn't block this. Except ssb-plugins, which needs to be merged: https://github.com/ssbc/ssb-server/pull/653

cli.js
```diff
@@ -0,0 +1,113 @@
...
+var minimist     = require('minimist')
+var muxrpcli     = require('muxrpcli')
+var cmdAliases   = require('./lib/cli-cmd-aliases')
+//var ProgressBar  = require('./lib/progress')
+var packageJson  = require('./package.json')
+
+//get config as cli options after --, options before that are
+//options to the command.
+
+module.exports = function (config, argv) {
```
ahdinosaur: can we name this function `cli` or something?

as down at the bottom, we call `module.exports(...)` and took me a few moments to grok what was happening.

cli.js
```diff
@@ -0,0 +1,113 @@
...
+    muxrpcli(argv, manifest, rpc, config.verbose)
+  })
+}
+
+if(!module.parent && process.title != 'browser') {
+  var argv = process.argv.slice(2)
+  var i = argv.indexOf('--')
+  var conf = argv.slice(i+1)
+  argv = ~i ? argv.slice(0, i) : argv
+  module.exports(Config(process.env.ssb_appname, minimist(conf)), argv)
```
ahdinosaur: there's _a lot_ happening here, but it's hard to tell _what_ is happening, what about:

```javascript
var args = process.argv.slice(2)
var [methodArgs, configArgs] = splitArgsBySeparator(args)
var appName = process.env.ssb_appname
var configOptions = minimist(configArgs)
var config = Config(appName, configOptions)

cli(config, methodArgs)

function splitArgvBySeparator (args) {
  var splitIndex = args.indexOf('--')
  var methodArgs = splitIndex == -1 ? args : args.slice(0, splitIndex)
  var configArgs = args.slice(splitIndex + 1)
  return [methodArgs, configArgs]
} 
```

cli.js
```diff
@@ -0,0 +1,113 @@
...
+      )
+      return
+    }
+
+    // run commandline flow
+    muxrpcli(argv, manifest, rpc, config.verbose)
+  })
+}
+
+if(!module.parent && process.title != 'browser') {
```
ahdinosaur: can there be a comment to describe what this is doing?

dominictarr:
@corlock had a problem that, running ssb-server inside of patchwork, couldn't use --help. I think whats happening is that --help is really, new, so that i'm not sure it's in the versions of modules that patchwork has?

PR #653, opened 2019-04-02
WIP: replace plugins/plugins with ssb-plugins

dominictarr:
not intending to merge this immediately... I think ssb-plugins module is only 90% finished. needs documentation at least, and maybe nicer way express wether something is out of process or not.

@cryptix what are your thoughts?


cryptix:
> what are your thoughts?

hmmm... long time since I looked into this and generally having a hard time judging _good enough_-ness in JS. I would have said it's still in tinker/proof-of-concept terreitory but I always feel bad and dirty when I write JS so maybe it's not that bad...

dominictarr:
I think it's good enough, except that I don't think it's good that the plugin decides wether it runs as in-process or out-of-process. that should be specified in the config. that means we need a new way to express plugins... that falls back to the old way. i guess it's plugins: `{<module>: boolean|string|{process: <boolean>?, name: <string>?},...}`

PR #674, opened 2019-05-27
refactor out compatibility tests

dominictarr:
this change removes the dependency testing script and uses `compatibility` for testing instead.
that does the same thing, but now it's a reusable tool!


christianbundy:
LGTM! Travis is still failing tests, but those have been broken since 14.1.10.

PR #685, opened 2019-08-22
add funding

mixmix:
see %FWwFYWPFAgQWh0fStot08GoDkK9c/HPYZvpE36vTKjs=.sha256

Notes:
- you only see an add **once** when you initially install, not when you run npm install later

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "cat-names": "^2.0.0",
     "compatibility": "^1.0.0",
     "dog-names": "~1.0.2",
+    "funding": "^1.0.3",
     "hexpp": "^2.0.0",
     "interleavings": "^1.0.0",
     "npm-install-package": "^2.1.0",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "ssb-server",
-  "version": "15.0.2",
+  "version": "15.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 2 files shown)


feross:
@mixmix, thanks for the support! It means a lot to me and I hope that having others join in will help shift the discussion on this topic, even if (like you said) we don't like sponsor messages as the ultimate funding model here.

Btw, I think you'll want to make this a dependency if you intend for end users to see the sponsor message. Otherwise, it will only be run for folks who are developing on `ssb-server`.

Garkazoid:
I am commenting here because you have shut down all other inbound forms of communication.

Attempting to create an issue on standard/standard where this conversation began results in the error:

"An owner of this repository has limited the ability to open an issue to users that have contributed to this repository in the past."

Attempting to reply to any other issue on standard/standard results in the error:

"An owner of this repository has limited the ability to comment to users that have contributed to this repository in the past."

On the https://github.com/feross/funding README page, is this line:
> Where can I provide feedback about this experiment?
>   You can open an issue. But please be kind. I'm a human with feelings. ❤️

Attempting to open an issue on feross/funding results in the error:
"An owner of this repository has limited the ability to open an issue from new users."

Attempting to comment on any other issue in feross/funding results in the error:
"An owner of this repository has limited the ability to comment from new users."


Having attempted to follow your explicit instructions for feedback and having found them all intentionally disabled, I leave my feedback here.

This is a terrible idea.

In no uncertain terms, what you are asking for the right to do is to include big, bold third-party banner advertising in a build diagnostic message channel.

It's clear that no consideration was given to the question of "what if everyone did this" for a couple reasons.  First, if everyone did this it would destroy the utility of build messages / build logs, which are important.  Second, there wasn't thought of distributing proceeds beyond Feross collecting cash and referrals for ads running in feross/funding and then Venmoing some proceeds out to his buddies if they include his adware in their build.

This is a terrible idea and you should listen to feedback rather than shutting down all channels of feedback.

package.json
```diff
@@ -75,6 +75,7 @@
     "cat-names": "^2.0.0",
     "compatibility": "^1.0.0",
     "dog-names": "~1.0.2",
+    "funding": "^1.0.3",
```
Garkazoid: Is it clear that Feross as an individual has squatted on the "funding" name in the npm global namespace to collect (and distribute?) all of the proceeds raised through distributing this adware?

dominictarr:
the funding experiment is officially over https://feross.org/funding-experiment-recap/
so closing this

PR #691, opened 2019-09-12
Fixes broken links in the docs

rabble:
The docs site has has broken links since Christian renamed it to ssb-server and it's been driving me crazy so i fixed it.


arj03:
Thanks!

PR #701, opened 2019-11-16
Bump ssb-keys to 7.2.0 for node v12 support

jedahan:
I used npm run reinstall, which changed a lot of the shrinkwrap.
Let me know if there was a better way.


jedahan:
There seems to be a circular dependency with the compatibility tests https://github.com/ssbc/ssb-replicate/pull/11 . Kinda annoying.

christianbundy:
I think this looks good, but I haven't been actively working with SSB-Server. What do you think @dominictarr?

dominictarr:
it's a minor so it should be fine, and there are tests, so run them and if it passes then publish.

arj03:
I just checked and this seems to have been fixed in 16 so closing.

PR #702, opened 2019-11-20
forward API documentation ref to ssb-db for up to date

Connoropolous:
I did a manual diff comparison between the two, and saw that in ssb-db only additions were made. There was nothing in this file that wasn't there, but the opposite was not true. There are new things there: 
This was the whole diff
Screen Shot 2019-11-20 at 10 17 17 AM

By the way, there wasn't even a link to this file in the README here, so I'm being even on the precocious side by even leaving this file.


christianbundy:
Nice! I think we can probably just rip this file out completely, what do you think? Happy to merge as-is or with a more aggressive deletion.

Context: I'm guessing that at one point we used this with [mdmanifest](https://github.com/ssb-junkyard/mdmanifest), which created a [muxrpc manifest file](https://github.com/ssbc/muxrpc#manifest) from a Markdown file like `api.md`.

Connoropolous:
Ok let's just delete, I agree

mixmix:
HOORAY!

Connoropolous:
Yeah, we are down to 3 competing API references only now :)

Connoropolous:
#705 is a discussion on how to drop that further, along with the issue in ssb-db that it links to

PR #703, opened 2019-11-20
Improve README organization and links

Connoropolous:
I separated example usage docs into two headers: 

## Starting an `ssb-server`
and
## Calling `ssb-server` Functions

because I find it confusing when they're mixed. Then, each has a Bash, and JS section. This helps clarify a lot of things.


christianbundy:
Love it, thank you!

PR #708, opened 2019-12-08
remove variables that are not used

mixmix:
I want to make code more accessible.
For me, I find I trip a lot on variables that aren't used - they take my attention away from what's important.
I think I might move through some of the core modules and remove or comment out variables which are not being used.

The intent of this is to make it more easy for more people to contribute to maintenance and growth <3

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,7 +10,3 @@ module.exports = createSsbServer()
 
 //this isn't really needed anymore.
 module.exports.createSsbServer = createSsbServer
-
-
-
-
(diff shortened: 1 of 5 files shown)


staltz:
:+1:

Did you use some automatic checker to detect those unused variables or did you do it by hand? Just want to make sure the unused variables are actually unused.

dominictarr:
I'm sure the _automatic checker_ was running the tests, right?

christianbundy:
The me was running the tests and the tests pass.

PR #722, opened 2020-05-21
add ssb-private1

mixmix:
with the upcoming `ssb-db@20` we're removing the classic private-box functionality by default.
To preserve the behaviour state of `ssb-server`, this adds it back into this distro.

**NOTE : this needs `ssb-db` to be upgraded (once that's published) before merging**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,6 @@ language: node_js
 sudo: false
 node_js:
   - 8
-  - 6
   - 10
+  - 12
+  - node
--- a/bin.js
+++ b/bin.js
@@ -43,6 +43,7 @@ if (argv[0] == 'start') {
   // import ssbServer and start the server
 
   var createSsbServer = require('./')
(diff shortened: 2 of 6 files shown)


mixmix:
2 errors about `getAddress` not being a known method ... not sure where that fell off the bandwagon. Having trouble logging stuff

christianbundy:
I think this is a regression that I caused in https://github.com/ssbc/ssb-db/pull/273

I didn't realize that `getAddress()` was one of those sneaky inheritance-based methods that appeared out of thin air. :/ I'll open a PR to add it back into SSB-DB.

staltz:
> NOTE : this needs ssb-db to be upgraded (once that's published) before merging

@mixmix could you convert this PR to a "Draft PR" so that we block accidental merges until ssb-db is upgraded?

christianbundy:
FWIW SSB-DB is upgraded, this branch + test fixes are available here: #724

PR #723, opened 2020-05-24
package.json: dep on sodium-native >=2.4.9

radfish:
sodium-native 2.4.2 does not build with nodejs 14.2.0 (compile errors reported during npm install of ssb-server).

Along with this patch, need to also update shrinkwrap and commit npm-shrinkwrap.json (I didn't do it, would rather a dev do it):

    npm install --save

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "pull-stream": "^3.6.2",
     "rimraf": "^2.4.2",
     "secret-stack": "^6.3.1",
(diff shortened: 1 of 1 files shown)


christianbundy:
Hey, thanks for this patch! Could you confirm that this is still a problem on v16.0.0 (latest)? It should work great on Node.js 8 - 14.

arj03:
I just checked and this seems to have been fixed in 16 so closing.

PR #727, opened 2020-06-11
Fix caps being interpreted as a Buffer

christianbundy:
Problem: The SSB-Server configures `caps` as a buffer, which breaks all
connections because it's supposed to be a base64 string.

Solution: Use the base64 string just like we're doing in SSB-Config.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@ var caps = require('ssb-caps')
 
 // create a sbot with default caps. these can be overridden again when you call create.
 function createSsbServer () {
-  return SecretStack({ caps: { shs: Buffer.from(caps.shs, 'base64') } })
+  return SecretStack({ caps })
     .use(require('ssb-db'))
 }
 module.exports = createSsbServer()


arj03:
Strange, when I was testing caps 3 months ago I didn't run into this problem. How did you find this?

christianbundy:
I just ran it from Node, `bin.js` worked fine but not when I used `require()`.

staltz:
Strange, this is secret-stack's responsibility, it should work either way because secret-stack converts it always to Buffer. See these lines: 

https://github.com/ssbc/secret-stack/blob/5a4767a817f6615a5bc256b8406e1460ce9ee9aa/plugins/shs.js#L47

and 

https://github.com/ssbc/secret-stack/blob/5a4767a817f6615a5bc256b8406e1460ce9ee9aa/plugins/shs.js#L12-L16

I don't doubt you experienced an issue, but I'm really curious why it broke with a buffer but not with a string.

clehner:
Tests pass locally. Thanks @christianbundy

PR #738, opened 2020-10-28
Bump ssb-links

clehner:
Fix crash as mentioned in https://github.com/ssbc/ssb-server/issues/737#issuecomment-718250217


arj03:
Shrinkwrap probably also needs update?

clehner:
@arj03 good catch, I've amended the commit

arj03:
Thanks

Issue #734, opened 2020-10-14
Not responding


clehner:

timjrobinson:
I encountered this again today in local dev after pulling in a lot of new data. Indexing was stalling at 97% and the server wasn't fully booting and so wasn't responding to whoami. 

After updating ssb-db to the latest version `20.4.0` with `npm i -s ssb-db@latest` it seems to be indexing correctly and responding to commands again. 

I did get some new errors in server logs that I hadn't seen before, so maybe they were causing the stall before:

```
could not retrive msg: Error [NotFoundError]: Key not found in database [@/3q5HcRfJu6KYA+zRmpwPEB5EN5QF8Ia4xhCrCTHtxw=.ed25519,1]
    at /home/tim/projects/ssb-server/node_modules/levelup/lib/levelup.js:188:15
    at /home/tim/projects/ssb-server/node_modules/encoding-down/index.js:75:21
could not retrive msg: Error [NotFoundError]: Key not found in database [@+sPe2KX1gMaYAiQJKTtBNlFT2bhZOHW07G7n1h4As+w=.ed25519,1]
    at /home/tim/projects/ssb-server/node_modules/levelup/lib/levelup.js:188:15
    at /home/tim/projects/ssb-server/node_modules/encoding-down/index.js:75:21
```

PR #739, opened 2020-11-09
Update v15

clehner:
Since #734, I am using and recommending ssb-server v15 over v16.

I would like to publish a new patch release of v15 with the following
changes:

- Update ssb-links to v3.0.10, as in #738
- Update chloride to v2.3.0, to bring in in sodium-native v3,
  [supporting Node v12+](https://github.com/ssb-js/chloride/pull/16)
- Backport #727 and #432, for good measure

There are no indexing conflicts that I am aware of between v15 and v16 currently - v16 just adds ssb-private1 - so it is safe to run both of these with the same `~/.ssb` directory (not at the same time, of course). Running v15 is useful for recovering from #734.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -8,7 +8,6 @@ var File         = require('pull-file')
 var explain      = require('explain-error')
 var Config       = require('ssb-config/inject')
 var Client       = require('ssb-client')
-var createHash   = require('multiblob/util').createHash
 var minimist     = require('minimist')
 var muxrpcli     = require('muxrpcli')
 var cmdAliases   = require('./lib/cli-cmd-aliases')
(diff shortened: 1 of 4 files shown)


ahdinosaur:
merging, as i'm keen to upgrade [`ssb-pub`](https://github.com/ahdinosaur/ssb-pub) with this release.

ref: https://github.com/ahdinosaur/ssb-pub/issues/37

ahdinosaur:
hm, i made a new version `15.3.0`, but not sure how to `npm publish`, `npm test` and specifically `compatibility` seems to always end on an error. i see https://github.com/ssbc/ssb-server/pull/628#issuecomment-459640976, but not sure how to resolve.

ahdinosaur:
yeah, seems like we're in a situation with:

> when two plugins have a dev dep that conflicts, the pretest script fails. a@2 a@3 would fail, it this case you will need to fix the plugins.

oh no, shaving yaks...

ahdinosaur:
well `compatibility` cycles between two sets of dev dependencies. both of them pass, so do i yolo and force publish? :smiley_cat: i'll make some lunch.

ahdinosaur:
published `ssb-server@15.3.0` :tada:

clehner:
Thanks @ahdinosaur

PR #762, opened 2022-03-25
Update installation instructions to work on Debian

mplorentz:
This updates the installation instructions in the README to work on a fresh installation of Ubuntu 20 LTS, and probably anything based on Debian. The current instructions assume that you have `curl` and `npm` already installed. They also assume you are running as a non-root user. I reordered the apt and nvm commands and added some clarification where I thought it was useful. 

I ran through these commands myself and was able to to generate and redeem invites at the end.


staltz:
Very good, thank you!

